PR #5, opened 2025-05-06
feat: enable random anti-fee sniping

### LINKED ISSUE #4: feature request: Sequence based anti-fee-sniping BIP326
Research possible solutions for implementing BIP326 in `bdk_tx` when creating PSBTs.

aagbotemi:
This PR implements Anti-Fee-Sniping with randomization as discussed in issue #4 

## Notes to the reviewers
The implementation adds randomization to anti-fee-sniping behavior:

1. Uses a 50/50 chance to choose between nLockTime and nSequence (when possible)
2. Adds a 10% chance to set either value further back in time (by a random value between 0-99)
3. Detects taproot inputs and their confirmation status

### Changed
- `CreatePsbtError::MissingFullTxForLegacyInput` and `CreatePsbtError::MissingFullTxForSegwitV0Input` now wrap `Input` in `Box` (breaking change)

### Added
- `PsbtParams::enable_anti_fee_sniping` field for BIP326 anti-fee-sniping protection
- `CreatePsbtError::InvalidLockTime` and `CreatePsbtError::UnsupportedVersion` error variants
- `Selection::create_psbt_with_rng` method for custom RNG

* [x] I've signed all my commits
* [x] I ran `cargo fmt` and `cargo clippy` before committing

Closes #4

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,20 +11,22 @@ license = "MIT OR Apache-2.0"
 readme = "README.md"
 
 [dependencies]
-miniscript = { version = "12", default-features = false }
+miniscript = { version = "12.3.5", default-features = false }
 bdk_coin_select = "0.4.0"
+rand_core = { version = "0.6.4", default-features = false }
+rand = { version = "0.8", optional = true }
 
 [dev-dependencies]
 anyhow = "1"
 bdk_tx = { path = "." }
-bitcoin = { version = "0.32", features = ["rand-std"] }
+bitcoin = { version = "0.32", default-features = false, features = ["rand-std"] }
 bdk_testenv = "0.13.0"
 bdk_bitcoind_rpc = "0.20.0"
 bdk_chain = { version = "0.23.0" }
 
 [features]
 default = ["std"]
-std = ["miniscript/std"]
+std = ["miniscript/std", "rand/std"]
 
 [[example]]
 name = "synopsis"
@@ -32,3 +34,6 @@ name = "synopsis"
 [[example]]
 name = "common"
 crate-type = ["lib"]
+
+[[example]]
+name = "anti_fee_sniping"
--- a/examples/anti_fee_sniping.rs
+++ b/examples/anti_fee_sniping.rs
@@ -1,0 +1,150 @@
+#![allow(dead_code)]
+use bdk_testenv::{bitcoincore_rpc::RpcApi, TestEnv};
+use bdk_tx::{
+    filter_unspendable_now, group_by_spk, selection_algorithm_lowest_fee_bnb, FeeStrategy, Output,
+    PsbtParams, ScriptSource, SelectorParams,
+};
+use bitcoin::{absolute::LockTime, key::Secp256k1, Amount, FeeRate, Sequence};
+use miniscript::Descriptor;
+
+mod common;
+
+use common::Wallet;
+
+fn main() -> anyhow::Result<()> {
+    let secp = Secp256k1::new();
+    let (external, _) = Descriptor::parse_descriptor(&secp, bdk_testenv::utils::DESCRIPTORS[0])?;
+    let (internal, _) = Descriptor::parse_descriptor(&secp, bdk_testenv::utils::DESCRIPTORS[1])?;
+
+    let env = TestEnv::new()?;
+    let genesis_hash = env.genesis_hash()?;
+    env.mine_blocks(101, None)?;
+
+    let mut wallet = Wallet::new(genesis_hash, external, internal.clone())?;
+    wallet.sync(&env)?;
+
+    let addr = wallet.next_address().expect("must derive address");
+
+    let txid1 = env.send(&addr, Amount::ONE_BTC)?;
+    env.mine_blocks(1, None)?;
+    wallet.sync(&env)?;
+    println!("Received confirmed input: {}", txid1);
+
+    let txid2 = env.send(&addr, Amount::ONE_BTC)?;
+    env.mine_blocks(1, None)?;
+    wallet.sync(&env)?;
+    println!("Received confirmed input: {}", txid2);
+
+    println!("Balance (confirmed): {}", wallet.balance());
+
+    let (tip_height, tip_time) = wallet.tip_info(env.rpc_client())?;
+    println!("Current height: {}", tip_height);
+    let longterm_feerate = FeeRate::from_sat_per_vb_unchecked(1);
+
+    let recipient_addr = env
+        .rpc_client()
+        .get_new_address(None, None)?
+        .assume_checked();
+
+    // When anti-fee-sniping is enabled, the transaction will either use nLockTime or nSequence.
+    //
+    // Locktime approach is used when:
+    // - RBF is disabled, OR
+    // - Any input requires locktime (non-taproot, unconfirmed, or >65535 confirmations), OR
+    // - There are no taproot inputs, OR
+    // - Random 50/50 coin flip chose locktime
+    //
+    // Sequence approach is used otherwise:
+    // - Sets tx.lock_time to ZERO
+    // - Modifies one randomly selected taproot input's sequence
+    //
+    // Once the approach is selected, to reduce transaction fingerprinting,
+    // - For nLockTime: With 10% probability, subtract a random 0-99 block offset from current height
+    // - For nSequence: With 10% probability, subtract a random 0-99 block offset (minimum value of 1)
+    //
+    // Note: When locktime is used, all sequence values remain unchanged.
+
+    let mut locktime_count = 0;
+    let mut sequence_count = 0;
+
+    for _ in 0..10 {
+        let selection = wallet
+            .all_candidates()
+            .regroup(group_by_spk())
+            .filter(filter_unspendable_now(tip_height, tip_time))
+            .into_selection(
+                selection_algorithm_lowest_fee_bnb(longterm_feerate, 100_000),
+                SelectorParams::new(
+                    FeeStrategy::FeeRate(FeeRate::from_sat_per_vb_unchecked(10)),
+                    vec![Output::with_script(
+                        recipient_addr.script_pubkey(),
+                        Amount::from_sat(50_000_000),
+                    )],
+                    ScriptSource::Descriptor(Box::new(internal.at_derivation_index(0)?)),
+                    wallet.change_policy(),
+                ),
+            )?;
+
+        let fallback_locktime: LockTime = LockTime::from_consensus(tip_height.to_consensus_u32());
+
+        let selection_inputs = selection.inputs.clone();
+
+        let psbt = selection.create_psbt(PsbtParams {
+            enable_anti_fee_sniping: true,
+            fallback_locktime,
+            fallback_sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+            ..Default::default()
+        })?;
+
+        let tx = psbt.unsigned_tx;
+
+        if tx.lock_time != LockTime::ZERO {
+            locktime_count += 1;
+            let locktime_value = tx.lock_time.to_consensus_u32();
+            let current_height = tip_height.to_consensus_u32();
+
+            let offset = current_height.saturating_sub(locktime_value);
+            if offset > 0 {
+                println!(
+                    "nLockTime = {} (tip height: {}, offset: -{})",
+                    locktime_value, current_height, offset
+                );
+            } else {
+                println!(
+                    "nLockTime = {} (tip height: {}, no offset)",
+                    locktime_value, current_height
+                );
+            }
+        } else {
+            sequence_count += 1;
+
+            for (i, inp) in tx.input.iter().enumerate() {
+                let sequence_value = inp.sequence.to_consensus_u32();
+
+                if (1..0xFFFFFFFD).contains(&sequence_value) {
+                    let input_confirmations = selection_inputs[i].confirmations(tip_height);
+                    let offset = input_confirmations.saturating_sub(sequence_value);
+
+                    if offset > 0 {
+                        println!(
+                            "nSequence[{}] = {} (confirmations: {}, offset: -{})",
+                            i, sequence_value, input_confirmations, offset
+                        );
+                    } else {
+                        println!(
+                            "nSequence[{}] = {} (confirmations: {}, no offset)",
+                            i, sequence_value, input_confirmations
+                        );
+                    }
+
+                    break;
+                }
+            }
+        }
+    }
+
+    println!("nLockTime approach used: {} times", locktime_count);
+    println!("nSequence approach used: {} times", sequence_count);
+
+    Ok(())
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,4 @@
 //! `bdk_tx`
-
-// FIXME: try to remove clippy "allows"
-#![allow(clippy::large_enum_variant)]
-#![allow(clippy::result_large_err)]
 #![warn(missing_docs)]
 #![no_std]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ mod rbf;
 mod selection;
 mod selector;
 mod signer;
+mod utils;
 
 pub use canonical_unspents::*;
 pub use finalizer::*;
@@ -30,6 +31,7 @@ pub use rbf::*;
 pub use selection::*;
 pub use selector::*;
 pub use signer::*;
+use utils::*;
 
 #[cfg(feature = "std")]
 pub(crate) mod collections {
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -1,12 +1,17 @@
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::fmt::{Debug, Display};
 
 use bdk_coin_select::FeeRate;
-use bitcoin::{absolute, transaction, Sequence};
 use miniscript::bitcoin;
+use miniscript::bitcoin::{
+    absolute::{self, LockTime},
+    transaction, Psbt, Sequence,
+};
 use miniscript::psbt::PsbtExt;
+use rand_core::RngCore;
 
-use crate::{Finalizer, Input, Output};
+use crate::{apply_anti_fee_sniping, Finalizer, Input, Output};
 
 const FALLBACK_SEQUENCE: bitcoin::Sequence = bitcoin::Sequence::ENABLE_LOCKTIME_NO_RBF;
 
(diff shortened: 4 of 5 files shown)


ValuedMammal:
Thanks @aagbotemi I'm still in the process of reviewing. I agree it would be nice to resolve the clippy error - my other suggestion would be to globally allow the lints in `lib.rs` for now and open an issue to fix them in a future PR, but in general I agree with your approach to simply `Box` the large enum variants.

aagbotemi:
> Thanks @aagbotemi I'm still in the process of reviewing.

Thank you @ValuedMammal. Clippy error and large enum variant error in the CI build has been fixed. For subsequent ones(if there is), I'll open another for it.

src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        let mut sequence_value = confirmation;
+
+        if random_probability(FURTHER_BACK_PROBABILITY) {
+            let random_offset = random_range(0, MAX_RANDOM_OFFSET);
+            sequence_value = sequence_value
+                .saturating_sub(random_offset)
+                .max(MIN_SEQUENCE_VALUE);
+        }
+
+        tx.input[*idx].sequence = Sequence(sequence_value);
```
ValuedMammal, line 90: To be clear, this seems to be making an assumption that the index found by enumerating the selection inputs corresponds to the same TxIn in `tx.input`, right?

ValuedMammal (commented):
I have some review notes

- For imports, prefer `alloc` over `std`. This way we can actually support no-std.
- I want to see a test of some kind that creates a PSBT with `enable_anti_fee_sniping` and checks the expected values of the locktime and/or sequence.

Cargo.toml
```diff
@@ -11,14 +11,15 @@ readme = "README.md"
 [dependencies]
 miniscript = { version = "12", default-features = false }
 bdk_coin_select = "0.4.0"
+rand_core = { version = "0.6.0", features = ["getrandom"] }
+bdk_chain = { version = "0.21" }
+bitcoin = { version = "0.32", features = ["rand-std"] }
```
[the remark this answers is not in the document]
aagbotemi: `bdk_chain` was made an explicit dependency, because I used it in the first commit before pr #1 was merged, and i had to refactor the code. I have added it to `dev-dependencies`

`bitcoin` dep has been made to have no default feature, and rust-bitcoin types have been imported via `miniscript`. though the types are still been imported from `bitcoin` in synopsis and common examples

src/selection.rs
```diff
@@ -53,6 +56,7 @@ impl Default for PsbtParams {
             fallback_locktime: absolute::LockTime::ZERO,
             fallback_sequence: FALLBACK_SEQUENCE,
             mandate_full_tx_for_segwit_v0: true,
+            enable_anti_fee_sniping: true,
```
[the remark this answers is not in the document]
aagbotemi: okay, that's true

src/selection.rs
```diff
@@ -146,8 +160,23 @@ impl Selection {
...
+            let height = params
+                .fallback_locktime
+                .is_block_height()
+                .then(|| params.fallback_locktime.to_consensus_u32())
+                .ok_or(CreatePsbtError::InvalidLockTime(params.fallback_locktime))?;
+
+            let current_height = bitcoin::absolute::Height::from_consensus(height)
+                .map_err(|_conversion_error| CreatePsbtError::InvalidHeight(height))?;
+
+            apply_anti_fee_sniping(&mut tx, &self.inputs, current_height, rbf_enabled);
```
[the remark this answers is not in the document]
aagbotemi: Yes. I have fixed that

src/utils.rs
aagbotemi: Yes, given that the length of `tx.input` is the same as `inputs`. I have now made a measure for correctness and safety - ensuring that there is right data with the right indices.

src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        "Probability must be between 0.0 and 1.0"
+    );
+
+    let mut rng = OsRng;
+    let rand_val = rng.next_u32() as f64;
+    let max_u32 = u32::MAX as f64;
+    (rand_val / max_u32) < probability
+}
+
+fn random_range(min: u32, max: u32) -> u32 {
```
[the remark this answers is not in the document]
aagbotemi: Yes, exactly, I was trying to avoid modulo bias, though it's very tiny. I will update the code with the simpler version.

Cargo.toml
```diff
@@ -32,3 +33,6 @@ name = "synopsis"
 [[example]]
 name = "common"
 crate-type = ["lib"]
+
+[[example]]
+name = "anti_fee_snipping"
```
ValuedMammal, line 38: Can we rename this to `anti_fee_sniping`.

src/selection.rs
```diff
@@ -63,13 +70,19 @@ pub enum CreatePsbtError {
     /// Attempted to mix locktime types.
     LockTypeMismatch,
     /// Missing tx for legacy input.
-    MissingFullTxForLegacyInput(Input),
+    MissingFullTxForLegacyInput(Box<Input>),
     /// Missing tx for segwit v0 input.
-    MissingFullTxForSegwitV0Input(Input),
+    MissingFullTxForSegwitV0Input(Box<Input>),
```
ValuedMammal, line 75: q: Should we change to `MissingFullTxForSegwitV0Input(OutPoint)` instead of having the `Box<Input>`?

src/utils.rs
```diff
@@ -0,0 +1,98 @@
...
+use rand_core::{OsRng, RngCore};
+
+/// Applies BIP326 anti‐fee‐sniping
+pub fn apply_anti_fee_sniping(
+    tx: &mut Transaction,
+    inputs: &[Input],
+    current_height: absolute::Height,
+    rbf_enabled: bool,
+) -> Result<(), CreatePsbtError> {
+    const MAX_SEQUENCE_VALUE: u32 = 65_535;
```
ValuedMammal, line 18: nit:
```suggestion
    const MAX_RELATIVE_HEIGHT: u32 = 65_535;
```

src/utils.rs
```diff
@@ -0,0 +1,98 @@
...
+    const USE_NLOCKTIME_PROBABILITY: u32 = 2;
+    const MIN_SEQUENCE_VALUE: u32 = 1;
+    const FURTHER_BACK_PROBABILITY: u32 = 10;
+    const MAX_RANDOM_OFFSET: u32 = 99;
```
ValuedMammal, line 22: Should this be `100` since we're doing the modulo?

src/utils.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // Check always‐locktime conditions
+    let must_use_locktime = inputs.iter().any(|input| {
+        let confirmation = input.confirmations(current_height);
+        confirmation == 0
+            || confirmation > MAX_SEQUENCE_VALUE
+            || !matches!(
+                input.plan().and_then(|plan| plan.witness_version()),
+                Some(WitnessVersion::V1)
+            )
```
ValuedMammal, line 50: I was thinking it might be simpler and more robust to test if the input is taproot by using `Script::is_p2tr`, rationale being that we always have a prev txout but may not have a plan.

```suggestion
            || !input.prev_txout().script_pubkey.is_p2tr()
```

src/utils.rs
```diff
@@ -0,0 +1,98 @@
...
+        // Use nLockTime
+        let mut locktime = current_height.to_consensus_u32();
+
+        if random_probability(&mut rng, FURTHER_BACK_PROBABILITY) {
+            let random_offset = random_range(&mut rng, MAX_RANDOM_OFFSET);
+            locktime = locktime.saturating_sub(random_offset);
+        }
+
+        let new_locktime = LockTime::from_height(locktime)
+            .map_err(|_| CreatePsbtError::InvalidHeight(locktime))?;
```
ValuedMammal, line 68: I think we can expect it to be a valid LockTime because `current_height` has the type `absolute::Height`. And probably just remove the `CreatePsbtError::InvalidHeight` variant.

```suggestion
        let new_locktime = LockTime::from_height(locktime).expect("must be valid Height");
```

ValuedMammal:
We should also be able to remove the clippy allow attributes in `lib.rs`.

src/utils.rs
ValuedMammal: Ok good point, the length must be the same or else we would be in the "use LockTime" branch.

Still, to not have the implicit ordering assumption I thought it would be better to find the correct input by outpoint, by maybe I'm over-analyzing.

<details>
  <summary>diff</summary>

```diff
-    let taproot_inputs: Vec<usize> = (0..tx.input.len())
-        .filter(|&idx| {
-            // Check if this input is taproot using the corresponding Input data
-            inputs
-                .get(idx)
-                .and_then(|input| input.plan())
-                .and_then(|plan| plan.witness_version())
-                .map(|version| version == WitnessVersion::V1)
-                .unwrap_or(false)
+    // vector of input_index and associated Input ref.
+    let taproot_inputs: Vec<(usize, &Input)> = tx
+        .input
+        .iter()
+        .enumerate()
+        .filter_map(|(vin, txin)| {
+            let input = inputs
+                .iter()
+                .find(|input| input.prev_outpoint() == txin.previous_output)?;
+            if input.prev_txout().script_pubkey.is_p2tr() {
+                Some((vin, input))
+            } else {
+                None
+            }
         })
         .collect();
 
     // Check always‐locktime conditions
@@ -70,10 +75,11 @@ pub fn apply_anti_fee_sniping(
         tx.lock_time = new_locktime;
     } else {
         // Use Sequence
         tx.lock_time = LockTime::ZERO;
-        let input_index = random_range(&mut rng, taproot_inputs.len() as u32) as usize;
-        let confirmation = inputs[input_index].confirmations(current_height);
+        let random_index = random_range(&mut rng, taproot_inputs.len() as u32);
+        let (input_index, input) = taproot_inputs[random_index as usize];
+        let confirmation = input.confirmations(current_height);
```
</details>

aagbotemi: You are absolutely correct. Your code is robust and helps the anti-fee-sniping logic to operate on the right inputs. 

Thank you for being thorough and helping the code to be correct and more maintainable.

src/selection.rs
aagbotemi: I agree it could be refactored.

Would you prefer I address this now or in a follow-up PR?

Cargo.toml
aagbotemi: Alright.

src/utils.rs
aagbotemi: Alright.

src/utils.rs
aagbotemi: Yes, it should be `100`. I have fixed it. Thank you.

src/utils.rs
aagbotemi: good, this approach is much better.

src/utils.rs
aagbotemi: Alright.

aagbotemi:
> I'm still in favor of using `Box` on `PlanOrPsbtInput` and `ScriptSource::Descriptor`.

Fixed! I've updated the code to use `Box`.

src/selection.rs
```diff
@@ -63,33 +71,43 @@ pub enum CreatePsbtError {
     /// Attempted to mix locktime types.
     LockTypeMismatch,
     /// Missing tx for legacy input.
-    MissingFullTxForLegacyInput(Input),
+    MissingFullTxForLegacyInput(OutPoint),
     /// Missing tx for segwit v0 input.
-    MissingFullTxForSegwitV0Input(Input),
+    MissingFullTxForSegwitV0Input(OutPoint),
```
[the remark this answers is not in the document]
aagbotemi: @ValuedMammal I have reverted the code.

src/utils.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn random_probability(rng: &mut OsRng, probability: u32) -> bool {
+    let rand_val = rng.next_u32();
+    rand_val % probability == 0
+}
+
+fn random_range(rng: &mut OsRng, max: u32) -> u32 {
+    rng.next_u32() % max
+}
```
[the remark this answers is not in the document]
aagbotemi: Yes, this is good. It helps to solve modulos bias

I would like to know your thought on using
```rs
use bitcoin::secp256k1::rand::Rng;

fn random_range(rng: &mut OsRng, max: u32) -> u32 {
    rng.gen_range(0..max)
}
```

aagbotemi:
> In general I think the git history would be cleaner if we didn't merge with merge commits but instead rebase the commits in this PR.

Alright, I've done the cleanup

ValuedMammal:
Approach ACK.

src/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+    let max = u32::MAX;
+    let max_multiple = max - (max % end);
+
+    loop {
+        let n = rng.next_u32();
+        if n < max_multiple {
+            return n % end;
+        }
+    }
+}
```
nymius: Here you are defining the largest contiguous unbiased region within the numeric range representable by the u32 type. From what I have discovered, `(-end) % end` is the preferred version in the latest literature, although the performance is almost the same. Probably the compiler optimizes both forms to the same instructions, but I would use `(-end) % (end)` to fingerprint the technique you're using here for future code readers.

src/utils.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    Ok(())
+}
+
+/// Returns true with probability 1/n.
+fn random_probability(rng: &mut OsRng, probability: u32) -> bool {
+    random_range(rng, probability) == 0
+}
+
+// Return a random value in the range [0, end].
```
nymius: You are trying to add a docstring here or just a comment? Use `///` for docstrings.

nymius (commented):
Approach ACK 4649e8dc76ac97e6228d1628892dc82fee5d9cc5.

I guess this will be squashed before merging, but I would reserve the `revert` type in commit messages for those actually reverting other commits, not manual commits undoing something.

I also checked the references in the BIP and bitcoin core seems to have implemented something similar back in the time.

aagbotemi:
> Approach ACK [4649e8d](https://github.com/bitcoindevkit/bdk-tx/commit/4649e8dc76ac97e6228d1628892dc82fee5d9cc5).
> 


Thank you for the review. I'm working on fixing them.

examples/anti_fee_sniping.rs
```diff
@@ -0,0 +1,137 @@
...
+    let tx = psbt.unsigned_tx;
+
+    // When anti-fee-sniping is enabled, the transaction will either use nLockTime or nSequence.
+    //
+    // Locktime approach is used when:
+    // - RBF is disabled, OR
+    // - Any input requires locktime (non-taproot, unconfirmed, or >65535 confirmations), OR
+    // - There are no taproot inputs, OR
+    // - Random 50/50 coin flip chose locktime
+    // - With 10% probability: subtracts random 0-99 block offset from current height
```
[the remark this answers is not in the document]
aagbotemi: Noted, will fix that. Thank you

src/lib.rs
```diff
@@ -1,8 +1,4 @@
 //! `bdk_tx`
-
-// FIXME: try to remove clippy "allows"
-#![allow(clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
nymius: But they were introduced in a previous different PR, and the implications are general to all the repository, that's why I recommend to move them to a previous separated commit.

src/utils.rs
```diff
@@ -0,0 +1,156 @@
...
+                .max(MIN_SEQUENCE_VALUE);
+        }
+
+        tx.input[input_index].sequence = Sequence(sequence_value);
+    }
+
+    Ok(())
+}
+
+/// Returns true with probability 1/n.
```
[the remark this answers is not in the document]
aagbotemi: The anti fee sniping implementation started from [here](https://github.com/bitcoindevkit/bdk/pull/1912#discussion_r2023587434). So, I have been using `rand_core` for the implementation since then. I can change to `rand` if we agree to use that.

cc: @ValuedMammal, we have access to using `Rng` trait using `rand`

examples/anti_fee_sniping.rs
```diff
@@ -0,0 +1,139 @@
...
+            "Value {} is out of range {}..={}",
+            height_val,
+            min_expected,
+            max_expected
+        );
+
+        if height_val >= min_expected && height_val <= max_expected {
+            println!("✓ Locktime is within expected range");
+        } else {
+            println!("⚠ Locktime is outside expected range");
```
nymius: Is this line reachable? I guess you want to show the feature is working, but if anti fee sniping works as intended, this is not going to ever happen. This isn't a test. Maybe showing the `nLockTime` or `nSequence` in a meaningful way is a better approach for the example.

nymius: Is this line reachable? I guess you want to show the feature is working, but if anti fee sniping works as intended, this is not going to ever happen. This isn't a test. Maybe showing the `nLockTime` or `nSequence` in a meaningful way is a better approach for the example.

examples/anti_fee_sniping.rs
```diff
@@ -0,0 +1,152 @@
...
+
+                    break;
+                }
+            }
+        }
+    }
+
+    println!("nLockTime approach used: {} times", locktime_count);
+    println!("nSequence approach used: {} times", sequence_count);
+    println!("Both approaches provide anti-fee-sniping protection:");
```
nymius: nit: this isn't providing new information, I recommend to remove it

Cargo.toml
```diff
@@ -13,11 +13,12 @@ readme = "README.md"
 [dependencies]
 miniscript = { version = "12", default-features = false }
 bdk_coin_select = "0.4.0"
```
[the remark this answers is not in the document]
ValuedMammal: Ideally we want to support no-std, so my question is how are we providing a source of randomness for no-std environments.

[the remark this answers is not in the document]
ValuedMammal: The approach we took for bdk_wallet is to depend on `rand_core` in order to have `RngCore` trait and essentially split the interface based on whether `std` feature is enabled. If `std`, then we have access to `rand` via `bitcoin::secp256k1::rand`. If no-std then the user can pass in something that implements `RngCore`.

```rust
/// Create PSBT.
#[cfg(feature = "std")]
pub fn create_psbt(&self, params: PsbtParams) -> Result<bitcoin::Psbt, CreatePsbtError> {
    self.create_psbt_with_rng(params, &mut rand::thread_rng())
}

/// Create PSBT with `rng`.
pub fn create_psbt_with_rng(&self, params: PsbtParams, rng: &mut impl RngCore) -> Result<bitcoin::Psbt, CreatePsbtError> { ... }

```

The other option is to feature-gate the `enable_anti_fee_sniping` option entirely but that seems a bit less useful.

aagbotemi:
> Sorry about the merge conflict.

Alright. I will rebase to resolve the conflict.

Cargo.toml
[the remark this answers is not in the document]
nymius: @aagbotemi could you adapt your code [here](https://github.com/bitcoindevkit/bdk-tx/commit/782c3e367fb772d9ea094609a6b03fa46346bf25#diff-23af5356f6001cd3993cd3c801fb7716ea02d1e504081b9fb569332db6107e80R142-R156) and the current state of the PR to the above solution? Using feature gates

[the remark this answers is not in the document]
aagbotemi: Alright, will do that.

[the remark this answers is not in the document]
nymius: I think you are missing the `std` implementation using `rand` full features.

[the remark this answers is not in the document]
nymius: > Looks good overall, not sure if any review comments are still unresolved.


@ValuedMammal I'm not sure we have fulfilled this requirements yet. Do you think we should roll a custom random range even if the user has `std` available?

ValuedMammal (approved):
ACK ed9df3cbe183b3bc5cd7c5e7eb66873a6cd628a7

- You could make a note somewhere that this also includes an API breaking change in `CreatePsbtError`
- Following up on the no-std discussion, we should have default features disabled for `bdk_coin_select` dependency, but I ran into an issue trying to implement it - bitcoindevkit/coin-select#36

nymius (commented):
ACK ed9df3cbe183b3bc5cd7c5e7eb66873a6cd628a7

A little bit redundant now, but worth vouching for regardless.

To complement @ValuedMammal, I think is enough to change commit ed9df3cbe183b3bc5cd7c5e7eb66873a6cd628a7 message from `feat: enable random anti-fee sniping` to `feat!: enable random anti-fee sniping` and comment about the rationale for the variant additions on the body of the commit message.

aagbotemi:
> make a note somewhere that this also includes an API breaking change in CreatePsbtError

Noted. Thank you.

> Following up on the no-std discussion, we should have default features disabled for bdk_coin_select dependency, but I ran into an issue trying to implement it - https://github.com/bitcoindevkit/coin-select/issues/36

no-std support is blocked by bdk_coin_select not having a no-std feature. I will fix it upstream.

>change commit https://github.com/bitcoindevkit/bdk-tx/commit/ed9df3cbe183b3bc5cd7c5e7eb66873a6cd628a7 message from feat: enable random anti-fee sniping to feat!: enable random anti-fee sniping and comment about the rationale for the variant additions on the body of the commit message.

Thank you for the suggestion. I will fix it

nymius (approved):
ACK 179860e9bbe4c763296d275d47462a192ca2a903

ValuedMammal:
ACK 179860e9bbe4c763296d275d47462a192ca2a903

src/selection.rs
```diff
@@ -52,6 +57,57 @@ pub struct PsbtParams {
...
+    ///         outputs: vec![], /* Outputs */
+    ///     };
+    ///     let psbt = selection.create_psbt(params)?;
+    ///     // the resulting transaction will have anti-fee-sniping applied.
+    ///     Ok(())
+    /// }
+    /// ```
+    ///
+    /// See [BIP326](https://github.com/bitcoin/bips/blob/master/bip-0326.mediawiki) for more details.
+    pub enable_anti_fee_sniping: bool,
```
evanlinjin, line 110: This parameter quietly expects `fallback_locktime` to be a block height, otherwise we get `InvalidLockTime`.

We should design our structures so that a height is always provided when we enable AFS so that `InvalidLockTime` is no longer needed. For example:

```rust
anti_fee_sniping: Option<AfsParams>,

struct AfsParams { pub tip_height: absolute::Height }
```

evanlinjin (commented):
The AFS logic operates on an already-built transaction, so it shouldn't be directly tied into `create_psbt`.

The only reason we have `create_psbt_with_rng` is to support this unnecessary coupling. It's also the only creator of the `InvalidLockTime` and `UnsupportedVersion` variants of `CreatePsbtError`.

### Suggested Shape

* `apply_anti_fee_sniping` already exists in the crate root, we need to make it public!
* Have a convenience method on `Selection`:

```rust
impl Selection {
    pub fn apply_anti_fee_sniping(
        &self,
        psbt: &mut Psbt,
        tip_height: absolute::Height,
        rng: &mut impl RngCore,
    ) -> Result<(), AntiFeeSnipingError> {
        apply_anti_fee_sniping(&mut psbt.unsigned_tx, &self.inputs, tip_height, rng)
    }
}
```

PR #6, opened 2025-05-14
Update `Cargo.toml` for initial release

evanlinjin:


ValuedMammal:
I noticed a few compile errors in `examples/common.rs` as a result of updating the dev-dependencies that I think can be easily addressed, other than that the changes to Cargo.toml look ok to me.

ValuedMammal:
I added ea61aff in order to appease CI, assuming it can be fixed in #5.

ValuedMammal (approved):
self-ACK d7eae73bc1e8c122e65d745cb2ad7810cb73213c

PR #9, opened 2025-05-30
ci: Add coverage.yml workflow

ValuedMammal:
A recent CI run can be found here https://github.com/bitcoindevkit/bdk-tx/actions/runs/15537664175/job/43740556358.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,55 @@
+name: Code Coverage
+
+on:
+  push:
+    branches: [
+      "master",
+      "test_ci"
+    ]
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  coverage:
+    name: Code Coverage
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: "-Cinstrument-coverage"
+      RUSTDOCFLAGS: "-Cinstrument-coverage"
+      LLVM_PROFILE_FILE: "./target/coverage/%p-%m.profraw"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install lcov tools
+        run: sudo apt-get install lcov -y
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@v1
+        with:
+            toolchain: stable
+            components: llvm-tools-preview
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.7.8
+      - name: Install grcov
+        run: if [[ ! -e ~/.cargo/bin/grcov ]]; then cargo install grcov; fi
+      - name: Test
+        run: cargo test --all-features
+      - name: Make coverage directory
+        run: mkdir coverage
+      - name: Run grcov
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -2,7 +2,10 @@ name: CI
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [
+      "master",
+      "test_ci",
+    ]
```
ValuedMammal, line 8: This is only here so I could test the branch in CI, and can probably be reverted.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.7.8
+      - name: Install grcov
+        run: if [[ ! -e ~/.cargo/bin/grcov ]]; then cargo install grcov; fi
+      - name: Test
+        run: cargo test --all-features
+      - name: Make coverage directory
+        run: mkdir coverage
+      - name: Run grcov
+        run: grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore '/*' --ignore '**/tests/**' --ignore '**/examples/**' -o ./coverage/lcov.info
```
nymius, line 44: Do you have in mind a second use of the `lcov` files other than source them to genhtml? Why not using `-t html`?

nymius (commented):
I have been testing grcov for `bdk-sp` and explored different combinations of flag to get the most of coverage reports:
- `--excl-start "#\[cfg\(test\)\]"`: In all the `bdk` repositories I have seen unit test are placed at the end of files, and not in the middle like some projects (correct me if I'm wrong). So I think this flag is useful in this case. I wouldn't use ` --excl-br-line "^\s*#\[cfg\(test\)\].*"` for these cases because I've experimented with it, and it ignores all `#[test]` tagged functions, but the utility functions inside the test module are not ignored, altering the final report numbers.
- `--excl-br-line "^\s*((debug_)?assert(_eq|_ne)?!|#\[derive\()"`: In this case I'm not so sure about the `debug_assert` and `assert` variations, as it may be used in some parts of the target code with actual implications in main logic of the program, but I don't have much of these in the code I have been using to test. In the case of `derive` I think is useful to exclude them in the case of default derivations. That code is boilerplate and cannot unless implemented manually, in which case the lines will be considered by the coverage report.
- `--excl-line  "^\s*\S+\s*(=>)?\s*unreachable!.*"`: I have some actually legit cases to use `unreachable`, so as long as the macro is not used lightly, I think is a good way to improve the signal of the report.

nymius:
After adding docstrings to [`bdk-sp` encoding module](https://github.com/bitcoindevkit/bdk-sp/commit/dd2ba1f3ac0327aebd247f8a94d13d4257f6c47d), `grcov` started to produce some false positives, so I started looking at other tools.
I ended up deciding with `cargo-llvm-cov` because it doesn't have those false positives, and also is able to ignore unit test code.
Although this is possible thanks to `nightly` attributes, as these tools are exclusively for the developing phase, and not release, I think it is OK to use `nightly` for these use cases.
`cargo-llvm-cov` also improves reports by adding branch coverage (thanks to `nightly`).
`--mdcd` gives [errors on `ci`](https://github.com/rust-lang/rust/issues/140677) so `--branch` is enough for now.

PR #15, opened 2025-07-23
Support for Child Pay For Parent (CPFP)

aagbotemi:
This PR introduces Child-Pays-For-Parent (CPFP) support to `bdk-tx` by implementing a new `CPFPSet` struct, `CPFPParams` and `cpfp_candidates` function. 
It also add  the ancestor fee calculation for CPFP.

This implementation supports:

- bumping multiple parent transaction with a single CPFP transaction
- Unconfirmed ancestors are considered in fee rate calculation
- Added `cpfp_candidates`
- Mempool compliance of ≤25 ancestors count

- [x] I've signed all my commits

Fixes #13


evanlinjin (changes requested):
Thank you for working on this!

Looking through the PR, I realized that I’m not entirely happy with the current state of `CanonicalUnspents`. Trying to build a CPFP API around it feels a bit clunky. I’m planning to replace this structure with `bdk_chain::CanonicalView` (WIP), which should make things cleaner.

To move forward with this PR, I suggest the following plan:

---

### Step 1 - Introduce standalone CPFP logic

Let’s first implement standalone CPFP logic that is decoupled from `CanonicalUnspents`.

```rust
pub struct CpfpParams {
    pub package_fee: Amount,
    pub package_weight: Weight,
    pub inputs: Vec<Input>,
    pub target_package_feerate: FeeRate,
    pub output_script: ScriptSource,
}

impl CpfpParams {
    fn into_selection(self) -> Result<Selection, CPFPError> {
        todo!("complete me")
    }
}
```

Here, `CpfpParams` assumes the caller has constructed it correctly. We should document this clearly, and later (in a separate PR) introduce helper methods to construct it more safely.

The implementation of `CpfpParams::into_selection will likely benefit from using `bdk_coin_select::CoinSelector` with [`select_all`](https://docs.rs/bdk_coin_select/latest/bdk_coin_select/struct.CoinSelector.html#method.select_all). You'll still need to do some math to determine the correct parameters. If you discover any missing functionality in `bdk_coin_select`, feel free to upstream improvements.

### Step 2 - Rebuild the example

Let's then rewrite `common::Wallet::create_cpfp_tx` using the `CpfpParams` logic. For simplicity, (since it's just an example), we can:
 
* Select the first owned output from each tx in `parent_txids`.
* Return an error if no owned outputs are found.

### Step 3 - Add tests

I recommend starting with a few simple tests:

* Given one original tx with feerate `x`, create CPFP tx with `target_package_feerate` of `x`. The resultant CPFP tx feerate should have feerate `x`.
* Given one original tx with feerate `x`, create a CPFP tx with `target_package_feerate` of `x + y`. The resulting package feerate should have a total feerate of `x + y` (`weight(original_tx + CPFP_tx) / fee(original_tx + CPFP_tx)`).

---

After this PR lands, we can start discussing ways to make the API less manual and more ergonomic.

aagbotemi:
> After this PR lands, we can start discussing ways to make the API less manual and more ergonomic.

Thank you for the review. I'll make the necessary fixes.

src/cpfp.rs
```diff
@@ -0,0 +1,203 @@
...
+            output_script,
+        }
+    }
+
+    /// Convert the CPFP parameters into selection.
+    ///
+    /// This method calculates the required child transaction fee to achieve the
+    /// target package feerate and creates a selection with the appropriate inputs
+    /// and outputs.
+    pub fn into_selection(self) -> Result<Selection, CpfpError> {
```
evanlinjin, line 55: Use the `CoinSelector` to calculate the weight. Everything else can be done with a simple formula:

```
cpfp_tx_fee = target_fee_rate * (package_weight + cpfp_tx_weight) - package_fee
```

evanlinjin (changes requested):

src/cpfp.rs
aagbotemi: Okay, I have made the fix

src/cpfp.rs
```diff
@@ -0,0 +1,203 @@
...
+        output_script: crate::ScriptSource,
+    ) -> Self {
+        Self {
+            package_fee,
+            package_weight,
+            inputs: inputs.into_iter().map(Into::into).collect(),
+            target_package_feerate,
+            output_script,
+        }
+    }
```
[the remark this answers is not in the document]
aagbotemi: Alright, I have fixed that

src/cpfp.rs
```diff
@@ -0,0 +1,115 @@
...
+/// This struct assumes the caller has constructed it correctly with:
+/// - `package_fee` accurately represents the total fees paid by all parent transactions
+/// - `package_weight` accurately represents the total weight of all parent transactions
+/// - `inputs` that reference valid, spendable UTXO
+/// - `target_package_feerate` is set to a value higher than the current effective
+///   package feerate (package_fee / package_weight) to make the CPFP effective
+/// - `output_script` that produces a valid output script
+///
+/// Violating these assumptions may result in errors during selection or invalid transactions.
+pub struct CpfpParams {
```
evanlinjin: For convenience, lets derive `Debug` and `Clone` here (at a minimum).

PR #18, opened 2025-08-08
feat(selector): allow multiple change sources with ChangeDescriptor

### LINKED ISSUE #17: Replace `change_descriptor: DefiniteDescriptor` field in `SelectorParams` by `change_spk: ScriptBuf`
**Description**
Replace `change_descriptor: DefiniteDescriptor` field in `SelectorParams` by `change_spk: ScriptBuf`.

https://github.com/bitcoindevkit/bdk-tx/blob/4f8a21350f532a249418c6e793755dea4f00a87d/src/selector.rs#L45

**Use case**

I have been using `bdk-tx` to wrap the selection and psbt creation together without doing much on `bdk-sp` side, but one of the things is keeping me from achieving this is the requirement of the `SelectorParams` to receive a `DefiniteDescriptor` for the change policy.

**Additional context**
In silent payments, descriptors are only useful to access keys, and wilcard derivation if used, is just to get the scan and spending keys respectively, after that no further derivation is performed on them.
For a silent payment wallet the conventional way to add a change output is to get an extra output sending to the same original code tweaked with the label `0` (at least that's the recommended way in BIP 352).
I have a way to provide a script pubkey to that parameter to replace later for the actual derived output, but working around the descriptors seems more cumbersome than just generalizing the requirement and requesting a script pubkey instead.

nymius:
**Description**

Previously, the `change_descriptor` field only accepted a definite descriptor to derive the change output script pubkey. This worked for most cases but failed for scenarios - such as silent payments - where a definite descriptor is not yet available.

To address this, introduce the `ChangeDescriptor` enum, which supports:

- The standard case of providing a definite descriptor.
- An alternative where the script pubkey and its maximum satisfaction weight can be specified directly.

This makes change output generation flexible enough to handle both standard and exceptional workflows.

Fixes #17 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [conventional commit guidelines](https://www.conventionalcommits.org/en/v1.0.0/)
* [x] I ran `cargo +nigthly fmt` before committing

#### New Features:

* [ ] I've added tests for the new feature

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -1,10 +1,10 @@
 use bdk_coin_select::{
     ChangePolicy, DrainWeights, InsufficientFunds, Replace, Target, TargetFee, TargetOutputs,
 };
-use bitcoin::{Amount, FeeRate, Transaction, TxOut, Weight};
+use bitcoin::{Amount, FeeRate, Transaction, Weight};
 use miniscript::bitcoin;
 
-use crate::{cs_feerate, DefiniteDescriptor, InputCandidates, InputGroup, Output, Selection};
+use crate::{cs_feerate, InputCandidates, InputGroup, Output, ScriptSource, Selection};
 use alloc::vec::Vec;
 use core::fmt;
 
@@ -15,7 +15,7 @@ pub struct Selector<'c> {
     target_outputs: Vec<Output>,
     target: Target,
     change_policy: bdk_coin_select::ChangePolicy,
-    change_descriptor: DefiniteDescriptor,
+    change_script: ScriptSource,
     inner: bdk_coin_select::CoinSelector<'c>,
 }
 
@@ -42,10 +42,13 @@ pub struct SelectorParams {
     /// To derive change output.
     ///
     /// Will error if this is unsatisfiable descriptor.
-    pub change_descriptor: DefiniteDescriptor,
+    pub change_script: ScriptSource,
 
     /// The policy to determine whether we create a change output.
     pub change_policy: ChangePolicyType,
+
+    /// Weight of the change output plus the future weight to spend the change
+    pub change_weight: DrainWeights,
 
     /// Params for replacing tx(s).
     pub replace: Option<RbfParams>,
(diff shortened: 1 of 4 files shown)


src/selector.rs
```diff
@@ -42,7 +83,7 @@ pub struct SelectorParams {
     /// To derive change output.
     ///
     /// Will error if this is unsatisfiable descriptor.
-    pub change_descriptor: DefiniteDescriptor,
+    pub change_descriptor: ChangeDescriptor,
```
ValuedMammal, line 86: Why not use `ScriptSource` here and change `SelectorParams` to account for the drain weights?

ValuedMammal: @nymius Will you look at this approach and let me know if it's a suitable solution? https://github.com/ValuedMammal/bdk-tx/commit/9080b30b48ed46d2803575095f813f48e5a1f8c2

ValuedMammal (approved):
ACK aadeca5e60deba5202d6069e6f4e7baae7fd83f4

evanlinjin:
ApproachNACK

There is a problem with the current approach: `SelectorParams` has two independent fields that can both specify the change output's satisfaction weight, and they can disagree.

- `change_script: ScriptSource` — when this is the `Descriptor` variant, the satisfaction weight is derivable from the descriptor via `max_weight_to_satisfy`.
- `change_weight: DrainWeights` — this also encodes the satisfaction weight (the spend cost of the change output).

The caller is responsible for keeping these in sync, but nothing enforces that. If `change_script` is a `Descriptor` and the caller provides a `DrainWeights` with a different satisfaction weight, we silently use the wrong value.

src/selector.rs
evanlinjin: > It just seems like the weight of eventually spending the change output should be more related to the change policy than the script source.

@ValuedMammal I think you are conflating two different things:

- **Change policy** - whether to create a change output (e.g. "no dust", "least waste"). That's a decision.
- **Satisfaction weight** - how much it costs to spend a particular script. That's a physical property of the script itself.

PR #24, opened 2025-10-22
Include `fallback_locktime` in locktime accumulation

ValuedMammal:
This allows the caller to specify an absolute locktime greater than may be required by a planned input.

### DIFF
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -31,9 +31,9 @@ pub struct PsbtParams {
 
     /// Fallback tx locktime.
     ///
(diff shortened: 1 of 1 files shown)


aagbotemi (commented):
tACK 6b76766fa1e389382af64a7f767441a59a6e2cd3

aagbotemi (commented):
Thank you for working on the test. I like the table-driven test approach here, it makes the different scenarios clear. I ran the test and it's working as expected.

Screenshot from 2025-11-12 00-06-02

PR #25, opened 2025-10-26
feat(selection)!: Add member `PsbtParams::sighash_type`

ValuedMammal:
This adds a new field `sighash_type` to `PsbtParams`. The option only applies to inputs that include a plan, as otherwise the given PSBT input can be expected to set a specific sighash type. The default is `None` which will not set an explicit sighash type for any input. In that case the sighash will typically cover all of the outputs.


aagbotemi (approved):
ACK f5cfb2a016dd11c47b760eb063b8a0d16925e0e3

PR #32, opened 2025-12-10
selector!: Swap `ChangePolicyType` for `bdk_coin_select::ChangePolicy`

### LINKED ISSUE #29: Make `ChangePolicyType` more flexible
`ChangePolicyType` is currently not very flexible, because it doesn't let you configure every parameter, for example the `min_value`. Further, `SelectorParams` has two different fields for `change_policy` and `change_weight` when these should really be treated together as part of the same policy when it comes to creating change, making the API awkward to work with.

ValuedMammal:
## Description

Fix #29 by removing `ChangePolicyType` from `SelectorParams` in favor of `bdk_coin_select::ChangePolicy`. The idea is that this affords maximum flexibility when defining the change policy.

Builds on #31 which implements absolute fee targeting. h/t aagbotemi. A test `test_absolute_fee_vs_feerate_target_value` is added that checks the expected `Target` is constructed based on the chosen `FeeStrategy`.

### Added

- selector: Added `FeeStrategy` enum which defines the fee target as either a feerate or fee amount.

### Changed

**BREAKING**

- `SelectorParams::target_feerate` field is changed to `fee_strategy`.
- `SelectorParams::change_policy` field is changed to have type `bdk_coin_select::ChangePolicy`.
- `SelectorParams::new` is changed to accept the `fee_strategy` and `change_policy` as inputs.

### Removed

- `SelectorParams::change_weight` field is removed now that the change weights are represented in the actual change policy.
- Removed `SelectorParams::to_cs_change_policy` as the conversion is no longer necessary.
- Removed `ChangePolicyType` enum to allow the user to construct the intended `ChangePolicy`.

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -14,7 +14,7 @@ pub struct Selector<'c> {
     candidates: &'c InputCandidates,
     target_outputs: Vec<Output>,
     target: Target,
-    change_policy: bdk_coin_select::ChangePolicy,
+    change_policy: ChangePolicy,
     change_script: ScriptSource,
     inner: bdk_coin_select::CoinSelector<'c>,
 }
@@ -45,8 +45,9 @@ pub struct SelectorParams {
     pub change_script: ScriptSource,
 
     /// The policy to determine whether we create a change output.
-    pub change_policy: ChangePolicyType,
+    pub change_policy: ChangePolicy,
 
+    // TODO: Remove this since the drain weights are now represented in the change policy.
     /// Weight of the change output plus the future weight to spend the change
     pub change_weight: DrainWeights,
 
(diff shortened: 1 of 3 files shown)


aagbotemi (approved):
Thank you for working on this PR. This approach is solid.

ACK b04db7eb8123065dd5b6f9919691517f8b94a63c

src/selector.rs
```diff
@@ -163,38 +159,31 @@ impl SelectorParams {
...
-                    dust_value,
-                    cs_feerate(self.target_feerate),
-                    cs_feerate(longterm_feerate),
-                )
+            FeeStrategy::AbsoluteFee(amount) => {
+                target_outputs.value_sum += amount.to_sat();
+                TargetFee {
+                    rate: bdk_coin_select::FeeRate::ZERO,
+                    replace: self.replace.as_ref().map(|r| r.to_cs_replace()),
+                }
```
evanlinjin, line 179: This looks like abusing the API. The underlying crate `bdk_coin_select` does not support coin selection purely with an absolute fee target.

This may result in unexpected behavior if paired with RBF. The `Replace` struct would still enforce incremental relay fee requirements, but the zero feerate means the selector wouldn't try to meet the original tx's feerate floor (`feerate_lb` is ignored). So you could end up with a replacement that satisfies BIP125 rule 4 (absolute fee) but not rule 6 (feerate), or vice versa.

If this feature is needed, we should upstream the changes.

PR #33, opened 2025-12-11
feat: add bip353 payment instructions support

sdmg15:
This PR adds BIP353 payment instructions support. The issue related to this PR can be found at https://github.com/bitcoindevkit/bdk_wallet/issues/13.

## Changes done:

- Added a feature flag `bip353` to conditionally enable the support
- Provides `resolve_dns_recipient(hrn, amount, network)`, this will resolve and validate the payment against what the resolution returned using a default resolver. The call to this function will return a `Payment` containing `scriptBuf`, `amount`, and `dnssec_proof` (which can be added to PSBT)
- Provides `resolve_dns_recipient_with_resolver(resolver, hrn, amount, network)` this will do the same as the previous function but this time using the user provided resolver.

## TODO

- Add tests
- Wrap the `ParseError` into an internal error type?
- Finish with implementing `resolve_dns_recipient_with_resolver`

## Questions

- The interfaces are async, should be provide synchronous one or let the end user decides what to do ?
- For FixedPaymentInstructions I was thinking of making the `amount` parameter as `Option` so that would mean the user is just ready to pay what the resolution returns, instead of him inputting a wrong amount. What do you think?

The functions added here are standalone. I'm happy to get guidance on how they could fit together with that is already implemented here

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 [dependencies]
 miniscript = { version = "12", default-features = false }
 bdk_coin_select = "0.4.0"
+bitcoin-payment-instructions = { version = "0.5.0", optional = true}
 
 [dev-dependencies]
 anyhow = "1"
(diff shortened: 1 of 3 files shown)


notmandatory:
I agree with @aagbotemi and I've added a related comment to the original issue here: https://github.com/bitcoindevkit/bdk_wallet/issues/13#issuecomment-3661474666

PR #38, opened 2026-01-31
Test MTP improvements

evanlinjin:
## Description

This PR provides the testing infrastructure and integration tests for #36 (MTP locktime fixes). The tests require unreleased upstream changes, so they're separated from the core fix.

### Dependency Updates
- Bump miniscript from 12.3.5 to 13.0.0
- Remove `Signer` wrapper type (KeyMap now implements GetKey directly)
- Adapt to bdk_chain `CanonicalView` API and `CheckPoint<Header>` generics
- Update Emitter for new mempool API (`update`/`evicted` fields)
- Patch dependencies to use unreleased branches with MTP support

### Test Infrastructure
- Convert repo to Cargo workspace with new `bdk_tx_testenv` crate under `testenv/`
- Move `Wallet` struct into `bdk_tx_testenv` so they can be used in both examples and tests
- Add `all_candidates_with(&Assets)` method for custom asset parameters

### Integration Tests
Add 4 time-based timelock tests that verify BDK predictions match Bitcoin Core's actual broadcast acceptance at exact boundaries:
- `test_absolute_time_timelock_logic`
- `test_relative_time_timelock_logic`
- `test_is_time_timelocked_absolute_unit`
- `test_is_time_timelocked_relative_unit`

### CI
- Remove `no-std` feature flag for miniscript (no longer exists in 13.0.0)

## Notes

- Based on #36
- Uses `[patch.crates-io]` to point to unreleased upstream branches


src/input.rs
```diff
@@ -360,39 +361,98 @@ impl Input {
...
+        let spending_height = tip_height
+            .to_consensus_u32()
+            .checked_add(1)
+            .expect("must not overflow");
+        if let Some(absolute::LockTime::Blocks(lt_height)) = self.plan.absolute_timelock() {
+            // Bitcoin Core's `IsFinalTx` uses strict less-than: a tx is final (unlocked) when
+            // `nLockTime < blockHeight`. This means `nLockTime = 100` is first spendable in
+            // block 101, not block 100. We return "locked" when the inverse is true.
+            return lt_height.to_consensus_u32() >= spending_height;
+        }
```
noahjoeris, line 375: This returns as soon as an absolute timelock exists, so it can skip a relative lock. E.g. `after(100) && older(10)` would return unlocked if `after(100)` is satisfied, even if `older(10)` is still active. 

Return early only when the absolute check is locked; otherwise check the relative lock too.

src/input.rs
```diff
@@ -360,39 +361,98 @@ impl Input {
...
+    /// Returns `None` if [`ConfirmationStatus::prev_mtp`] is required but unavailable.
+    ///
+    /// `tip_mtp` is `MTP(tip)`, or `MTP(spending_block - 1)`, as per BIP-0068.
+    pub fn is_time_timelocked(&self, tip_mtp: absolute::Time) -> Option<bool> {
+        if let Some(absolute::LockTime::Seconds(lt_time)) = self.plan.absolute_timelock() {
+            // Bitcoin Core's `IsFinalTx` (with BIP 113) uses strict less-than: a tx is final
+            // (unlocked) when `nLockTime < MTP`. This means `nLockTime = T` is first spendable
+            // when `MTP > T`, not when `MTP == T`. We return "locked" when the inverse is true.
+            return Some(lt_time.to_consensus_u32() >= tip_mtp.to_consensus_u32());
+        }
```
noahjoeris, line 403: same here. see comment on `is_block_timelocked`

tests/timelock.rs
```diff
@@ -0,0 +1,1280 @@
...
+        .get_new_address(None, None)?
+        .address()?
+        .assume_checked();
+
+    let selection = wallet
+        .all_candidates_with(&assets)
+        .regroup(group_by_spk())
+        .filter(filter_unspendable_now(tip_height, Some(tip_mtp)))
+        .into_selection(
+            selection_algorithm_lowest_fee_bnb(FeeRate::from_sat_per_vb_unchecked(1), 100_000),
```
noahjoeris, line 383: Deprecated. See https://github.com/bitcoindevkit/bdk-tx/pull/51
```suggestion
            selection_algorithm_lowest_fee_bnb(FeeRate::from_sat_per_vb(1).expect("valid fee rate"), 100_000),
```

src/input.rs
```diff
@@ -482,23 +548,60 @@ impl InputGroup {
...
-            .any(|input| input.is_timelocked(tip_height, tip_time))
+            .any(|input| input.is_block_timelocked(tip_height))
+    }
+
+    /// Whether any contained input is locked by a time-based timelock (absolute or relative).
+    ///
+    /// `tip_mtp` is `MTP(tip)`, or `MTP(spending_block - 1)`, as per BIP-0068.
+    pub fn is_time_timelocked(&self, tip_mtp: absolute::Time) -> Option<bool> {
+        for input in &self.0 {
+            if input.is_time_timelocked(tip_mtp)? {
```
noahjoeris, line 568: The group result should not depend on input order.

Expected behavior:
- Return `Some(true)` if any input is definitely locked.
- Return `Some(false)` if all inputs are definitely unlocked.
- Return `None` only if no input is definitely locked, but at least one input is unknown.

Current behavior returns `None` on the first unknown input, so `[unknown, locked]` returns `None` instead of `Some(true)`

src/input.rs
```diff
@@ -482,23 +548,60 @@ impl InputGroup {
...
+    /// Whether any contained input is locked by any timelock constraint.
+    ///
+    /// `tip_mtp` is `MTP(tip)`, or `MTP(spending_block - 1)`, as per BIP-0068.
+    pub fn is_timelocked(
+        &self,
+        tip_height: absolute::Height,
+        tip_mtp: Option<absolute::Time>,
+    ) -> Option<bool> {
+        for input in &self.0 {
+            if input.is_timelocked(tip_height, tip_mtp)? {
```
noahjoeris, line 584: Same here. see comment above (line 568)

src/input.rs
```diff
@@ -482,23 +548,60 @@ impl InputGroup {
...
-            .all(|input| input.is_spendable_now(tip_height, tip_time))
+    ///
+    /// `tip_mtp` is `MTP(tip)`, or `MTP(spending_block - 1)`, as per BIP-0068.
+    pub fn is_spendable(
+        &self,
+        tip_height: absolute::Height,
+        tip_mtp: Option<absolute::Time>,
+    ) -> Option<bool> {
+        for input in &self.0 {
+            if !input.is_spendable(tip_height, tip_mtp)? {
```
noahjoeris, line 600: Same here. See comments above.

noahjoeris (commented):
Nice one, this looks good overall. I left a few comments on bugs that should be fixed before merge.

PR #39, opened 2026-02-17
refactor!: Remove FeeStrategy add ChangeScript and improve SelectorParams

### LINKED ISSUE #40: SelectorParams has redundant satisfaction weight fields that can silently disagree
## Problem

`SelectorParams` currently has two independent fields that can both specify the change output's satisfaction weight:

- `change_script: ScriptSource` — when this is the `Descriptor` variant, the satisfaction weight is derivable from the descriptor via `max_weight_to_satisfy`.
- `change_policy: ChangePolicy` (which is `bdk_coin_select::ChangePolicy`) — also encodes the satisfaction weight via `DrainWeights::spend_weight`.

The caller is responsible for keeping these in sync, but nothing enforces that. If `change_script` is a `Descriptor` and the caller provides a `ChangePolicy` with a different `spend_weight`, we silently use the wrong value.

For the `ScriptSource::Script` case (e.g. silent payments), we genuinely need the caller to provide the satisfaction weight externally since it can't be derived. But for the `Descriptor` case, it should be derived automatically — having the caller also specify it manually is redundant and error-prone.

Additionally, `FeeStrategy::AbsoluteFee` was implemented by smuggling the fee into `value_sum` and setting the feerate to zero, which made the coin selector unaware of the actual fee and interacted poorly with RBF logic.

## Proposed solution

- Introduce `ChangeScript` that bundles a raw script with an optional `satisfaction_weight`, or holds a `DefiniteDescriptor` from which it is derived automatically. `DrainWeights` is then computed internally — single source of truth.
- Replace the raw `bdk_coin_select::ChangePolicy` with a `ChangePolicy` enum (`NoDust`, `NoDustLeastWaste`) that is converted to the `bdk_coin_select` type internally.
- Replace `FeeStrategy` with a direct `target_feerate: FeeRate` field.

## Context

- https://github.com/bitcoindevkit/bdk-tx/pull/18#issuecomment-3906226448
- https://github.com/bitcoindevkit/bdk-tx/pull/32#discussion_r2815577014

### LINKED ISSUE #42: Revert `FeeStrategy`
> FeeStrategy::AbsoluteFee was implemented by smuggling the fee into value_sum and setting the feerate to zero. This meant the coin selector had no awareness of the fee, which interacted poorly with RBF logic (the zero feerate bypasses the original tx's feerate floor). Removed in favor of a direct target_feerate: FeeRate field.

Originally posted by @evanlinjin https://github.com/bitcoindevkit/bdk-tx/pull/39#issue-3953812998

evanlinjin:
Fixes #40 
Fixes #42

## Summary

This PR fixes two issues with `SelectorParams`:

**1. Redundant satisfaction weight (invalid representation)**

`SelectorParams` previously had two independent fields that could both specify the change output's satisfaction weight:

- `change_script: ScriptSource` — when this is the `Descriptor` variant, the satisfaction weight is derivable from the descriptor via `max_weight_to_satisfy`.
- `change_policy: ChangePolicy` — also encodes the satisfaction weight (the spend cost of the change output).

Nothing enforced that these agreed with each other, so callers could silently provide inconsistent values.

Fixed by introducing **`ChangeScript`**, which bundles a raw script with an optional `satisfaction_weight`, or holds a `DefiniteDescriptor` from which it is derived automatically. `DrainWeights` is now computed internally from `ChangeScript` — there is a single source of truth.

`ChangeScript::Descriptor` also accepts optional `satisfaction_assets` — when provided, the satisfaction weight is computed via `Plan` for a tighter estimate (useful for multisig/complex descriptors). Otherwise it falls back to `max_weight_to_satisfy`.

The raw `bdk_coin_select::ChangePolicy` is also replaced with a **`ChangePolicy` enum** (`NoDust`, `NoDustLeastWaste`) that is converted to the `bdk_coin_select` type internally. This is `Debug + Clone + PartialEq + Eq`.

Both `ChangeScript` and `ChangePolicy` have constructor methods to reduce boilerplate:
- `ChangeScript::from_descriptor(desc)`, `from_descriptor_with_assets(desc, assets)`, `from_script(script, weight)`
- `ChangePolicy::no_dust()`, `no_dust_least_waste(rate)`, with a builder-style `.min_value(amt)`

**2. Hacky `AbsoluteFee` implementation**

`FeeStrategy::AbsoluteFee` was implemented by smuggling the fee into `value_sum` and setting the feerate to zero. This meant the coin selector had no awareness of the fee, which interacted poorly with RBF logic (the zero feerate bypasses the original tx's feerate floor). Removed in favor of a direct `target_feerate: FeeRate` field.

### Design rationale

The satisfaction weight is a physical property of the script — it describes how much it costs to spend a particular output. It is not a policy choice. When `change_script` is a descriptor, the satisfaction weight is literally derived from it. The fact that it *affects* coin selection doesn't make it *part of* the change policy, any more than the dust threshold is (and dust is already derived from the script).

For the raw script case (e.g. silent payments), the satisfaction weight can't be derived and must be provided externally — but it's still a property of the script, not a policy decision. Bundling it into `ChangeScript::Script` makes this explicit.

## Context

* https://github.com/bitcoindevkit/bdk-tx/pull/18#issuecomment-3906226448
* https://github.com/bitcoindevkit/bdk-tx/pull/32#discussion_r2815577014

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -28,7 +28,7 @@ pub struct Selector<'c> {
 /// * Error on anything that does not satisfy mempool policy.
 ///   If the caller wants to create non-mempool-policy conforming txs, they can just fill in the
 ///   fields directly.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct SelectorParams {
     /// Target feerate.
     ///
@@ -62,27 +62,98 @@ pub struct SelectorParams {
 /// For a [`DefiniteDescriptor`], the satisfaction weight is derived automatically. For a raw
 /// script (e.g. silent payments), the caller may provide it. It can be omitted if the change
 /// policy does not require waste calculations.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug)]
 pub enum ChangeScript {
     /// A raw script pubkey.
     Script {
         /// The output script.
         script: ScriptBuf,
-        /// Weight needed to satisfy this script in a future spending transaction.
+        /// The weight of the witness/scriptSig data needed to spend this script in a future
+        /// transaction.
+        ///
+        /// This is the same value as
+        /// [`Plan::satisfaction_weight`](miniscript::plan::Plan::satisfaction_weight) and is used
+        /// by coin selection to estimate the cost of spending the change output.
         ///
         /// Can be `None` if the change policy does not require waste calculations.
         satisfaction_weight: Option<Weight>,
     },
     /// A definite descriptor from which the script and satisfaction weight are both derived.
-    Descriptor(Box<DefiniteDescriptor>),
+    Descriptor {
+        /// The descriptor.
+        descriptor: Box<DefiniteDescriptor>,
+        /// Assets available for satisfying the descriptor.
+        ///
+        /// If provided, the satisfaction weight is computed via [`Plan`](miniscript::plan::Plan)
+        /// for a tighter estimate. If `None`, falls back to
+        /// [`max_weight_to_satisfy`](DefiniteDescriptor::max_weight_to_satisfy).
+        satisfaction_assets: Option<miniscript::plan::Assets>,
+    },
 }
 
 impl ChangeScript {
+    /// Create from a [`DefiniteDescriptor`].
+    ///
+    /// The satisfaction weight is derived via
+    /// [`max_weight_to_satisfy`](DefiniteDescriptor::max_weight_to_satisfy).
+    pub fn from_descriptor(descriptor: DefiniteDescriptor) -> Self {
+        Self::Descriptor {
+            descriptor: Box::new(descriptor),
+            satisfaction_assets: None,
+        }
+    }
+
+    /// Create from a [`DefiniteDescriptor`] with known assets.
+    ///
+    /// The satisfaction weight is derived via [`Plan`](miniscript::plan::Plan) for a tighter
+    /// estimate based on the provided assets.
+    pub fn from_descriptor_with_assets(
+        descriptor: DefiniteDescriptor,
+        assets: miniscript::plan::Assets,
+    ) -> Self {
+        Self::Descriptor {
+            descriptor: Box::new(descriptor),
+            satisfaction_assets: Some(assets),
+        }
+    }
+
+    /// Create from a raw script.
+    pub fn from_script(script: ScriptBuf, satisfaction_weight: Option<Weight>) -> Self {
+        Self::Script {
+            script,
+            satisfaction_weight,
+        }
+    }
+
     /// Convert to a [`ScriptSource`], discarding the satisfaction weight.
     pub fn source(&self) -> ScriptSource {
         match self {
             ChangeScript::Script { script, .. } => ScriptSource::Script(script.clone()),
-            ChangeScript::Descriptor(descriptor) => ScriptSource::Descriptor(descriptor.clone()),
+            ChangeScript::Descriptor { descriptor, .. } => {
+                ScriptSource::Descriptor(descriptor.clone())
+            }
+        }
+    }
+
+    fn satisfaction_weight(&self) -> Result<Weight, SelectorError> {
+        match &self {
+            ChangeScript::Script {
+                satisfaction_weight,
+                ..
+            } => satisfaction_weight.ok_or(SelectorError::MissingSatisfactionWeight),
+            ChangeScript::Descriptor {
+                descriptor,
+                satisfaction_assets,
+            } => match satisfaction_assets {
+                Some(assets) => descriptor
+                    .clone()
+                    .plan(assets)
+                    .map(|p| Weight::from_wu_usize(p.satisfaction_weight()))
+                    .map_err(|_| SelectorError::InsufficientAssets),
+                None => descriptor
+                    .max_weight_to_satisfy()
+                    .map_err(SelectorError::Miniscript),
+            },
         }
     }
 }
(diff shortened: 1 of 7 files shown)


aagbotemi (commented):
Since the previous implementation for `AbsoluteFee` was broken for RBF, removing it is correct. Targeting absolute fee is a legitimate use case and the open PR on [bdk_coin_select #38](https://github.com/bitcoindevkit/coin-select/pull/38) adding `TargetFee::absolute` as a field addresses this. Once that merges, `SelectorParams` and `to_cs_target()` should be updated by exposing and mapping onto it.

src/selector.rs
```diff
@@ -25,44 +28,184 @@ pub struct Selector<'c> {
...
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct SelectorParams {
-    /// Fee targeting strategy.
+    /// Target feerate.
     ///
-    /// Either target a specific feerate or an absolute fee.
-    pub fee_strategy: FeeStrategy,
+    /// The actual feerate of the resulting transaction may be higher due to RBF requirements or
+    /// rounding.
```
nymius, line 36: Can't this be considered external to the target feerate?
When including the target feerate, can't the feerate be computed externally based on RBF rules?

nymius (commented):
cACK ffd1211338aa1abc0b17fd73863b33efd53e1193

This is similar to #18 original's idea.

I have to review closer the RBF considerations, but I understand the removal of `FeeStrategy` as it is mainly needed for the `absolute fee` case, and it can be covered in other better ways.

I agree that `ChangePolicy` is a decision based on constraints, and the weight derived from `ChangeScript` is a property, so I don't see any concern raising from clearly reflecting that in the API.

ValuedMammal:
Appreciate your efforts, I'd just ask that you stick to the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md#contribution-workflow) in particular with regard to the contribution workflow.

- Can you link to the issue being fixed, or open one if you haven't already
- If you do a PR can you try to fix one issue at a time
- Can you include a test if adding a feature or fixing a bug

evanlinjin:
@ValuedMammal 

> Can you link to the issue being fixed, or open one if you haven't already

Added: https://github.com/bitcoindevkit/bdk-tx/issues/40

> If you do a PR can you try to fix one issue at a time

These changes can't be split — they're one refactor of `SelectorParams`:

- `ChangeScript` fixes the redundant satisfaction weight.
- `ChangePolicy` enum is required by the fix — the old `bdk_coin_select::ChangePolicy` bundled `DrainWeights` which contained the satisfaction weight. Can't remove the redundancy without replacing this type.
- Removing `FeeStrategy::AbsoluteFee` is entangled with the same `to_cs_target()` being refactored.
- Constructor methods are just helpers for the new types.

Splitting would create intermediate states where the API is half-migrated.

> Can you include a test if adding a feature or fixing a bug

You can't really "test" that two fields can't disagree anymore - the fix is that the problematic API shape no longer exists. The compiler enforces it.

src/selector.rs
evanlinjin: `target_feerate` is a lower bound, not an exact target. The actual feerate can be higher for multiple reasons — RBF feerate floor is one, but coin selection also raises it when dropping the change output reduces waste. The doc comment is just stating this fact.

Making the caller compute the RBF-adjusted feerate externally would just push complexity outward for no gain — the library already has `RbfParams` with the original tx feerates, so it can enforce the floor automatically.

src/selector.rs
```diff
@@ -159,31 +303,71 @@ impl SelectorParams {
...
-                }
-            }
+            spend_weight: if self.change_policy.considers_waste() {
+                // This code assumes that the change spend transaction is segwit.
+                bitcoin::TxIn::default().segwit_weight().to_wu()
+                    + self.change_script.satisfaction_weight()?.to_wu()
+            } else {
+                // Spend weight is not needed.
+                0
+            },
```
ValuedMammal: 🔧  It'd be better to not assume the spending transaction is segwit, unless the change script specifically encodes a witness program. You could make it a configurable `is_segwit` flag as part of the ChangePolicy.

🔧  Change outputs don't typically have a spend weight of 0. Recall that the cost of change is property of the output script, which has implications for coin selection, and less so with the ChangePolicy enum. So I'd say remove the `considers_waste` function.

src/selector.rs
```diff
@@ -25,44 +28,184 @@ pub struct Selector<'c> {
...
+            }
+        }
+    }
+
+    fn satisfaction_weight(&self) -> Result<Weight, SelectorError> {
+        match &self {
+            ChangeScript::Script {
+                satisfaction_weight,
+                ..
+            } => satisfaction_weight.ok_or(SelectorError::MissingSatisfactionWeight),
```
ValuedMammal: We're getting the satisfaction weight from the user, so just make it non-optional rather than introduce a new error path.

ValuedMammal:
From what I gather your two primary concerns are to

1. Revert FeeStrategy
2. Change the shape of the API

ValuedMammal (changes requested):

src/selector.rs
evanlinjin: The satisfaction weight may not always be easy to compute for raw scripts (e.g. silent payments or exotic script types). Making it non-optional forces callers to compute it even when their `ChangePolicy` doesn't need it. The error path exists precisely to catch the mismatch at runtime rather than burden all callers upfront.

src/selector.rs
evanlinjin: > Change outputs don't typically have a spend weight of 0.

Yes, but we aren't using that value at this point anyway.

evanlinjin: The `spend_weight: 0` for `NoDust` is intentional — `bdk_coin_select::ChangePolicy::min_value` doesn't use spend weight in its calculations. The reason we gate on `considers_waste()` is that `ChangeScript::from_script` allows `satisfaction_weight: None`, so callers using `NoDust` with a raw script don't need to provide a satisfaction weight they'll never use. Always computing the real spend weight would force every caller to provide one, making the API more burdensome for no practical benefit.

Regarding the segwit assumption — agreed it's not ideal, but whether the future spending tx is segwit depends on ALL its inputs, not just the change script. We can't know that at coin selection time. That said, assuming segwit is a conservative estimate (slightly overestimates spend cost by ~1 WU), so it seems like a reasonable default.

src/selector.rs
evanlinjin: @nymius can you explain why change output's spend weight cannot be determined for a silent payment tx?

Aren't silent payment outputs always P2TR keyspend path? If this is the case, we can remove the `Option<spend weight>` for `ChangeScript` as @ValuedMammal suggested and have a cleaner API.

Edit: I could also be misinterpreting this comment here: https://github.com/bitcoindevkit/bdk-tx/pull/18#discussion_r2337042509

nymius: > Aren't silent payment outputs always P2TR keyspend path?

Yes, that's true and naturally, it also applies to silent payment change outputs.

> Edit: I could also be misinterpreting this comment here: #18 (comment)

I must have eluded the almost fixed size of p2tr keypath spend witnesses there, and if not, I was giving more importance to the +-1 WU difference than I should have.

aagbotemi (approved):
ACK 8189b0317d722365185d7ae4d7ef4406f8c4312c

src/selector.rs
```diff
@@ -25,44 +28,142 @@ pub struct Selector<'c> {
...
+    ///
+    /// The satisfaction weight (cost of spending the change output in the future) is derived from
+    /// this. For descriptors it is computed automatically; for raw scripts it must be provided.
+    pub change_script: ChangeScript,
+
+    /// Dust relay feerate used to calculate the dust threshold for change outputs.
     ///
-    /// Will error if this is unsatisfiable descriptor.
-    pub change_script: ScriptSource,
+    /// If `None`, defaults to 3 sat/vB (the Bitcoin Core default for `-dustrelayfee`).
```
ValuedMammal, line 50:
```rs
/// If `None` the dust threshold is set by the [`minimal_non_dust`] value of the change script.
```

ValuedMammal (approved):
ACK 8189b0317d722365185d7ae4d7ef4406f8c4312c

PR #43, opened 2026-03-21
fix(selection): Improve handling of fallback locktime

ValuedMammal:
## Description

This is a follow-up to #24 which correctly implements the fallback locktime during PSBT creation, while enabling the fallback to contribute to the maximum accumulated locktime, provided that the lock type units are in agreement.

### Notes to the Reviewers 

Previously when setting the locktime of the transaction the fallback locktime was being chained onto the input locktimes causing a `LockTypeMismatch` error whenever a height-based fallback conflicted with a time-based input CLTV requirement. Since the `PsbtParams::default` defines a fallback_locktime that is height based, this made the default fallback incompatible with a time based input requirement.

This is fixed so that the fallback is only used if no inputs require an absolute timelock, or if the fallback is the same unit, then the maximum locktime value is used.

Added a test to check that the fallback is ignored if the lock type is incompatible.

### DIFF
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -174,16 +174,26 @@ impl core::fmt::Display for CreatePsbtError {
 impl std::error::Error for CreatePsbtError {}
 
 impl Selection {
-    /// Returns none if there is a mismatch of units in `locktimes`.
-    fn _accumulate_max_locktime(
+    /// Accumulates the maximum locktime from an iterator of input-required locktimes.
+    ///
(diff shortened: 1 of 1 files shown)


ValuedMammal:
In the last push

- Changed `fn _accumulate_max_locktime` to `accumulate_max_locktime` (dropped the leading `_`)
- `accumulate_max_locktime` now returns `CreatePsbtError::LockTypeMismatch` in case there is a mismatch of lock type units.

aagbotemi (approved):
ACK 260a46c0811c6efdc2668c2c6b95c8b8917223dc

PR #44, opened 2026-04-05
fix(finalizer): set finalized to false on error and add full coverage

noahjoeris:
## Description
This PR fixes a bug in Finalizer::finalize() where finalized remained true after an input finalization error, for example when a signature is missing. That could cause PSBT output metadata to be cleared even though finalization did not fully succeed.
I also added tests with full coverage.

### DIFF
--- a/src/finalizer.rs
+++ b/src/finalizer.rs
@@ -161,3 +161,269 @@ impl FinalizeMap {
         self.0
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


ValuedMammal:
Why not use `is_finalized` on the FinalizeMap directly to check if the outputs can be cleared and do away with the `finalized` variable altogether?

noahjoeris:
yep makes sense. Done.

ValuedMammal (approved):
ACK cfdad3d8eaec80fa1612f66786dec307d4dcaf3c

PR #45, opened 2026-04-08
ci: add code coverage workflow

### LINKED ISSUE #8: [ci] Add code coverage workflow
See example coverage workflow here https://github.com/bitcoindevkit/bdk/blob/066c4cd160f6d4c70409b299a173f31f93728fae/.github/workflows/code_coverage.yml

noahjoeris:
# Description
This PR adds CI code coverage workflow. I tried to be as close to the bdk flow as possible. Adapted minor things to this repo.

Closes #8

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -1,0 +1,49 @@
+name: Code Coverage
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  Coverage:
+    name: Code Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: nightly-2026-04-09
+          override: true
+          cache: true
+          components: llvm-tools-preview
+      - name: Install cargo-llvm-cov
+        run: if [[ ! -e ~/.cargo/bin/cargo-llvm-cov ]]; then cargo install cargo-llvm-cov; fi
+      - name: Make coverage directory
+        run: mkdir coverage
+      - name: Test and export lcov coverage
+        run: cargo llvm-cov -q --branch --all --ignore-filename-regex "(examples/|tests/)" --all-features --lcov --output-path ./coverage/lcov.info
+        env:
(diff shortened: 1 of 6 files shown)


noahjoeris:
I just saw the issue linked to an older bdk `code_coverage.yml`. I'll try to update. -> Done

ValuedMammal:
Thank you @noahjoeris just noting that I think there are plans to switch to codecov-action for uploading coverage reports using BDK's `CODECOV_TOKEN` (bitcoindevkit/bdk#2144), maybe we can try that here?

noahjoeris:
Changed it to codecov. You might still need to set the codecov token tho :)

ValuedMammal:
The token exists at the org level but maybe not activated for this repo yet. While I look into it here's some additional notes

- We can probably exclude unit/integration test modules from coverage as well
- I'm not sure what --doctests is doing and seems to have produced a warning but I didn't investigate
- It would be ideal to try cargo llvm-cov with a more recent version of the nightly compiler
- Unrelated, but we should also using caching in rust.yml which is made easy by actions-rust-lang/setup-rust-toolchain

noahjoeris:
Good points thx. I updated.

ValuedMammal:
Ah I think merging to master is all that's needed for codecov.io.

ValuedMammal:
Can we add this? https://github.com/ValuedMammal/bdk-tx/commit/8d2ea5b1bc941bcd68730471830683c6c76db461

noahjoeris:
Cool, I added it. Seems to work and exclude unit tests. Though I get this warn.
```
warning: feature `coverage_attribute` is declared but not used
```
cargo llvm-cov compiles the crate twice, once as the library (no cfg(test)), once as the test binary. In the library pass, #[cfg(test)] mod tests is stripped before attribute checking, so rustc never sees #[coverage(off)] being used and flags the feature as unused. In the test pass it's seen and works as intended.

We probably don't want to just silence it with: `#![cfg_attr(coverage_nightly, allow(unused_features))]`. Can we live with that warning?

ValuedMammal (approved):
ACK cb52b8cf499cf8fc765a5431d66573e3e8482f25

ValuedMammal:
> In the library pass, #[cfg(test)] mod tests is stripped before attribute checking, so rustc never sees #[coverage(off)] being used and flags the feature as unused.

I tried to reproduce the warning in `bdk_wallet` with no success. I suppose one fix could be to use coverage(off) for a non-test item like a Display impl.

PR #47, opened 2026-04-14
Add configurable `max_weight` to `SelectorParams`

### LINKED ISSUE #41: Implement builder interface for SelectorParams.
Building transactions through `SelectorParams` currently gets no feedback if they accidentally create dust outputs or include multiple `OP_RETURN` outputs. These mistakes only surface at relay time with opaque node rejections.
Implement a builder interface for `SelectorParams` that validates target outputs against configurable mempool standardness rules, cathching dust recipients, multiple `OP_RETURN` outputs, oversized `OP_RETURN` scripts, and missing required fields at construction time.

**Proposed solution**
Introduce a `SelectorParamsBuilder` as a separate type that validates and produces `SelectorParams`.
The `SelectorParams::builder()` chains methods and call `.build()` which validates against a configurable `MempoolPolicy` struct.

**Context**
See [selector.rs L20–L27](https://github.com/bitcoindevkit/bdk-tx/blob/338857fa892345bda610076d644d950b7f1b97d1/src/selector.rs#L20-L27)

### LINKED ISSUE #49: Enforce standardness rules for TxBuilder::add_data
~~"TxBuilder::add_data could enforce common standardness rules (max 80 bytes and at most one such output per transaction). It would also prevent a potential panic to be triggered from within rust-bitcoin (in push_slice_no_opt) if the data is too large, although i don't think it's realistic as a request for creating a 4GiB OP_RETURN would probably fail before then."~~

The policy for OP_RETURN has changed since [Bitcoin Core v30.0](https://bitcoincore.org/en/releases/30.0/)+, per release notes:

"-datacarriersize is increased to 100,000 by default, which effectively uncaps the limit (as the maximum transaction size limit will be hit first). It can be overridden with -datacarriersize=83 to revert to the limit enforced in previous versions. ([#32406](https://github.com/bitcoin/bitcoin/pull/32406))

Multiple data carrier (OP_RETURN) outputs in a transaction are now permitted for relay and mining. The -datacarriersize limit applies to the aggregate size of the scriptPubKeys across all such outputs in a transaction, not including the scriptPubKey size itself. ([#32406](https://github.com/bitcoin/bitcoin/pull/32406))"

aagbotemi:
### Description
Adds a configurable `SelectorParams::max_weight` that bounds the estimated weight of the transaction.

Closes #41 
Closes #49

### Note for reviewers
- **Why `try_finalize` checks weight, and before the target.** During selection the change output hasn't been decided, so the code can only estimate the transaction's weight without it. `try_finalize` assembles the final transaction, so it knows the real weight including change and checks that against the cap, the exact check, not a duplicate of the selection-time estimate. Weight is checked before the target so a selection that covers the value but is over-weight reports `MaxWeightExceeded` rather than `CannotMeetTarget`.

- **Why the error types were reshaped.** A weight failure used to surface as a value error, so the caller couldn't tell "not enough coins" from "coins too heavy to fit"; `SelectorError` now splits these into `InsufficientFunds` and `MaxWeightExceeded`. `MaxWeightExceeded` is a unit variant since `try_finalize` already reports the weights. And since `try_finalize` now returns `Result` instead of `Option`, `CannotMeetTarget` folds into `SelectorError` too, so the old `CannotMeetTarget` struct and `IntoSelectionError::CannotMeetTarget` variant are gone and every selector-side failure flows through the one `Selector(SelectorError)` arm.

- **Why the cap is stored on the target, not on `Selector`.** `to_cs_target()` writes `SelectorParams::max_weight` into `Target::max_weight`, so there's one source of truth, `Selector::max_weight()` reads it back.

- **Dependency.** Builds on bdk_coin_select#48 (`Target::max_weight`); the `Cargo.toml` git dependency is temporary and reverts to a release before merge.

### Changelog notice
- Added `max_weight` field to `SelectorParams`.
- Added `Selector::max_weight`, `Selector::weight_with` and `Selector::weight`.
- `Selector::try_finalize` now returns `Result<Selection, SelectorError>` and can fail with the new `SelectorError::MaxWeightExceeded` or `SelectorError::CannotMeetTarget`.
- `Selector::select_until_target_met` returns `Result<(), SelectorError>` and can now fail with `SelectorError::MaxWeightExceeded` or `SelectorError::InsufficientFunds`.
- `SelectorError::MaxWeightExceeded` is now a unit variant; added `SelectorError::InsufficientFunds`.
- Removed `CannotMeetTarget` struct and `IntoSelectionError::CannotMeetTarget`. `SelectorError::CannotMeetTarget` is now a variant.

### Checklists
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I've added tests for the new feature
* [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/selector.rs
+++ b/src/selector.rs
@@ -1,4 +1,4 @@
-use bdk_coin_select::{InsufficientFunds, Replace, Target, TargetFee, TargetOutputs};
+use bdk_coin_select::{Drain, Replace, SelectError, Target, TargetFee, TargetOutputs};
 use bitcoin::{Amount, FeeRate, ScriptBuf, Transaction, Weight};
 use miniscript::bitcoin;
 
@@ -23,12 +23,8 @@ pub struct Selector<'c> {
 
 /// Parameters for creating tx.
 ///
-/// TODO: Create a builder interface on this that does checks. I.e.
-/// * Error if recipient is dust.
-/// * Error on multi OP_RETURN outputs.
-/// * Error on anything that does not satisfy mempool policy.
-///   If the caller wants to create non-mempool-policy conforming txs, they can just fill in the
-///   fields directly.
+/// Required fields are set via [`SelectorParams::new`]; optional fields are
+/// set directly on the struct.
 #[derive(Debug)]
 pub struct SelectorParams {
     /// Target feerate.
@@ -64,6 +60,11 @@ pub struct SelectorParams {
 
     /// Params for replacing tx(s).
     pub replace: Option<RbfParams>,
+
+    /// Maximum allowed weight of the transaction.
+    ///
+    /// Defaults to the consensus block-weight limit ([`Weight::MAX_BLOCK`]).
+    pub max_weight: Weight,
 }
 
 /// Source of the change output script and its spending cost.
@@ -247,7 +248,7 @@ impl RbfParams {
 }
 
 impl SelectorParams {
-    /// With default params.
+    /// Construct params from the required fields.
     pub fn new(
         target_feerate: FeeRate,
         target_outputs: Vec<Output>,
@@ -261,6 +262,7 @@ impl SelectorParams {
             change_longterm_feerate: None,
             replace: None,
             change_dust_relay_feerate: None,
+            max_weight: Weight::MAX_BLOCK,
         }
     }
 
@@ -274,12 +276,14 @@ impl SelectorParams {
             fee: TargetFee {
                 rate: self.target_feerate.max(feerate_lb).into_cs_feerate(),
                 replace: self.replace.as_ref().map(|r| r.to_cs_replace()),
+                absolute: 0,
             },
             outputs: TargetOutputs::fund_outputs(
                 self.target_outputs
                     .iter()
                     .map(|o| (o.txout().weight().to_wu(), o.value.to_sat())),
             ),
+            max_weight: Some(self.max_weight.to_wu()),
         }
     }
 
@@ -331,29 +335,13 @@ impl SelectorParams {
     }
 }
 
-/// Error when the selection is impossible with the input candidates
-#[derive(Debug)]
-pub struct CannotMeetTarget;
-
-impl fmt::Display for CannotMeetTarget {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "meeting the target is not possible with the input candidates"
-        )
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for CannotMeetTarget {}
-
 /// Selector error
 #[derive(Debug)]
 pub enum SelectorError {
     /// Miniscript error (e.g. the change descriptor is inherently unsatisfiable).
     Miniscript(miniscript::Error),
     /// Meeting the target is not possible with the input candidates.
-    CannotMeetTarget(CannotMeetTarget),
+    CannotMeetTarget,
     /// The provided assets cannot satisfy the change descriptor.
     InsufficientAssets,
     /// Input candidates have absolute timelocks of mixed units (some height-based, others
@@ -363,19 +351,40 @@ pub enum SelectorError {
     /// Filter the [`InputCandidates`] down to a single-unit subset before constructing the
     /// [`Selector`].
     LockTypeMismatch,
+    /// The selection exceeds the maximum allowed transaction weight.
+    MaxWeightExceeded,
+    /// Not enough value in the candidates to meet the target.
+    InsufficientFunds(bdk_coin_select::InsufficientFunds),
 }
 
 impl fmt::Display for SelectorError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Miniscript(err) => write!(f, "{err}"),
-            Self::CannotMeetTarget(err) => write!(f, "{err}"),
+            Self::CannotMeetTarget => write!(
+                f,
+                "meeting the target is not possible with the input candidates"
+            ),
             Self::InsufficientAssets => {
                 write!(f, "provided assets cannot satisfy the change descriptor")
             }
             Self::LockTypeMismatch => {
                 write!(f, "input candidates have absolute timelocks of mixed units")
             }
+            Self::MaxWeightExceeded => write!(
+                f,
+                "selection exceeds the maximum allowed transaction weight"
+            ),
+            Self::InsufficientFunds(e) => write!(f, "{}", e),
+        }
+    }
+}
+
+impl From<SelectError> for SelectorError {
+    fn from(e: SelectError) -> Self {
+        match e {
+            SelectError::InsufficientFunds(e) => Self::InsufficientFunds(e),
+            SelectError::MaxWeightExceeded => Self::MaxWeightExceeded,
         }
     }
 }
@@ -400,7 +409,7 @@ impl<'c> Selector<'c> {
         let change_script = params.change_script.source();
 
         if target.value() > candidates.groups().map(|grp| grp.value().to_sat()).sum() {
-            return Err(SelectorError::CannotMeetTarget(CannotMeetTarget));
+            return Err(SelectorError::CannotMeetTarget);
         }
 
         // Verify that all inputs agree on absolute timelock unit (height vs time).
@@ -465,8 +474,9 @@ impl<'c> Selector<'c> {
     }
 
     /// Select in order until target is met.
-    pub fn select_until_target_met(&mut self) -> Result<(), InsufficientFunds> {
-        self.inner.select_until_target_met(self.target)
+    pub fn select_until_target_met(&mut self) -> Result<(), SelectorError> {
+        self.inner.select_until_target_met(self.target)?;
+        Ok(())
     }
 
     /// Whether we added the change output to the selection.
@@ -483,14 +493,38 @@ impl<'c> Selector<'c> {
         Some(has_drain)
     }
 
+    /// Maximum allowed transaction weight.
+    pub fn max_weight(&self) -> Option<Weight> {
+        self.target.max_weight.map(Weight::from_wu)
+    }
+
+    /// Estimated weight of the transaction for a given drain decision.
+    fn weight_with(&self, drain: Drain) -> Weight {
+        Weight::from_wu(self.inner.weight(self.target.outputs, drain.weights))
+    }
+
+    /// Estimated weight of the transaction.
+    pub fn weight(&self) -> Weight {
+        let drain = self.inner.drain(self.target, self.change_policy);
+        self.weight_with(drain)
+    }
+
     /// Try get final selection.
     ///
-    /// Return `None` if target is not met yet.
-    pub fn try_finalize(&self) -> Option<Selection> {
-        if !self.inner.is_target_met(self.target) {
-            return None;
+    /// # Errors
+    ///
+    /// - [`SelectorError::CannotMeetTarget`] if the target is not met yet.
+    /// - [`SelectorError::MaxWeightExceeded`] if the estimated transaction weight exceeds [`SelectorParams::max_weight`].
+    pub fn try_finalize(&self) -> Result<Selection, SelectorError> {
+        let drain = self.inner.drain(self.target, self.change_policy);
+        if let Some(max_wu) = self.target.max_weight {
+            if self.weight_with(drain).to_wu() > max_wu {
+                return Err(SelectorError::MaxWeightExceeded);
+            }
         }
-        let maybe_change = self.inner.drain(self.target, self.change_policy);
+        if !self.inner.is_target_met(self.target) {
+            return Err(SelectorError::CannotMeetTarget);
+        }
         let to_apply = self.candidates.groups().collect::<Vec<_>>();
         let inputs = self
             .inner
@@ -500,13 +534,13 @@ impl<'c> Selector<'c> {
             .cloned()
             .collect();
         let mut outputs = self.target_outputs.clone();
-        if maybe_change.is_some() {
+        if drain.is_some() {
             outputs.push(Output::from((
                 self.change_script.clone(),
-                Amount::from_sat(maybe_change.value),
+                Amount::from_sat(drain.value),
             )));
         }
-        Some(Selection::new(inputs, outputs))
+        Ok(Selection::new(inputs, outputs))
     }
 }
 
@@ -558,6 +592,34 @@ mod tests {
             Selector::new(&candidates, params),
             Err(SelectorError::LockTypeMismatch)
         ));
+        Ok(())
+    }
+
+    #[test]
+    fn into_selection_errors_when_max_weight_exceeded() -> anyhow::Result<()> {
+        let input = setup_cltv_input(absolute::LockTime::from_consensus(10_000))?;
+        let recipient_spk = input.prev_txout().script_pubkey.clone();
+        let candidates = InputCandidates::new([], [input]);
+
+        let mut params = SelectorParams::new(
+            FeeRate::from_sat_per_vb_u32(2),
+            vec![Output::with_script(recipient_spk, Amount::from_sat(10_000))],
+            ChangeScript::from_script(ScriptBuf::new(), Weight::ZERO),
+        );
+        params.max_weight = Weight::from_wu(100);
+
+        let err = candidates
+            .into_selection(|selector| selector.select_until_target_met(), params)
+            .unwrap_err();
+
+        assert!(
+            matches!(
+                err,
+                IntoSelectionError::SelectionAlgorithm(SelectorError::MaxWeightExceeded)
+            ),
+            "expected MaxWeightExceeded, got {err:?}"
+        );
+
         Ok(())
     }
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,8 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 # TODO: coin-select dependency should set no default features
-bdk_coin_select = { version = "0.4.1" }
+# Temporary: unreleased MaxWeight metric. Replace with a released bdk_coin_select version before merge.
+bdk_coin_select = { git = "https://github.com/aagbotemi/coin-select", branch = "feat/max-weight-metric" }
 miniscript = { version = "12.3.7", default-features = false }
 rand_core = { version = "0.6.4", default-features = false }
 rand = { version = "0.8", optional = true }
--- a/examples/synopsis.rs
+++ b/examples/synopsis.rs
@@ -3,7 +3,7 @@ use bdk_tx::{
     filter_unspendable, group_by_spk, selection_algorithm_lowest_fee_bnb, Output, PsbtParams,
     SelectorParams, Signer,
 };
-use bitcoin::{key::Secp256k1, Amount, FeeRate};
+use bitcoin::{key::Secp256k1, Amount, FeeRate, Weight};
 use miniscript::Descriptor;
 
 mod common;
@@ -143,6 +143,7 @@ fn main() -> anyhow::Result<()> {
                     change_dust_relay_feerate: None,
                     // This ensures that we satisfy mempool-replacement policy rules 4 and 6.
                     replace: Some(rbf_params),
+                    max_weight: Weight::MAX_BLOCK,
                 },
             )?;
 
--- a/src/input_candidates.rs
+++ b/src/input_candidates.rs
@@ -6,10 +6,7 @@ use bitcoin::{absolute, FeeRate, OutPoint};
 use miniscript::bitcoin;
 
 use crate::collections::{BTreeMap, HashSet};
-use crate::{
-    CannotMeetTarget, FeeRateExt, Input, InputGroup, Selection, Selector, SelectorError,
-    SelectorParams,
-};
+use crate::{FeeRateExt, Input, InputGroup, Selection, Selector, SelectorError, SelectorParams};
 
 /// Input candidates.
 #[must_use]
@@ -207,7 +204,7 @@ impl InputCandidates {
             .map_err(IntoSelectionError::SelectionAlgorithm)?;
         let selection = selector
             .try_finalize()
-            .ok_or(IntoSelectionError::CannotMeetTarget(CannotMeetTarget))?;
+            .map_err(IntoSelectionError::Selector)?;
         Ok(selection)
     }
 }
@@ -219,8 +216,6 @@ pub enum IntoSelectionError<E> {
     Selector(SelectorError),
     /// Selection algorithm failed.
     SelectionAlgorithm(E),
-    /// The target cannot be met
-    CannotMeetTarget(CannotMeetTarget),
 }
 
 impl<E: fmt::Display> fmt::Display for IntoSelectionError<E> {
@@ -232,7 +227,6 @@ impl<E: fmt::Display> fmt::Display for IntoSelectionError<E> {
             IntoSelectionError::SelectionAlgorithm(error) => {
                 write!(f, "selection algorithm failed: {error}")
             }
-            IntoSelectionError::CannotMeetTarget(error) => write!(f, "{error}"),
         }
     }
 }
@@ -249,6 +243,7 @@ pub fn selection_algorithm_lowest_fee_bnb(
     move |selector| {
         let target = selector.target();
         let change_policy = selector.cs_change_policy();
+
         selector
             .inner_mut()
             .run_bnb(


noahjoeris (commented):
I think the policy checks should be integrated more directly into the normal transaction-building flow.
Right now the caller has to remember to do a separate post-build check:

```rust
let policy = MempoolPolicy { tip_height, tip_mtp };
policy.check_all(&selection, &tx)?;
```

This is easy for callers to forget. It also asks them to provide both a Selection and a Transaction, even though the transaction should be derived from Selection + PsbtParams. A checked path should instead construct the transaction itself and validate exactly the transaction the crate is going to return.

Suggested shape:

- Output checks in `SelectorParamsBuilder`
  - `build()` could validate against a default policy.
  - `build_with_policy(policy)` could validate against a custom `MempoolPolicy`.
  - For unchecked, e.g. `build_unchecked()`

- Final transaction checks should happen in `Selection`.
  - `create_psbt(...)`  defaults to e.g. Core v30
  - Add `create_psbt_with_policy(policy)` and `create_psbt_unchecked()`
  - This avoids a public `check_all(&Selection, &Transaction)` API and makes `SelectionTxMismatch` unnecessary

- So I think we should have some form of abstraction/config over policy.
  - No hardcoded policy values but make them part of MempoolPolicy struct.
  - Also chain-tip data is an unrelated runtime input that should be passed alongside, not embedded
  - Use `MempoolPolicy::bitcoin_core_v30()` as the default, for example.
  - Users should be able to pass a different policy if their node has different relay settings.
  - Then this policy can be passed into both `SelectorParamsBuilder::build_with_policy(...)` and `Selection::create_psbt_with_policy(...)`.

Let me know what you think.

aagbotemi:
Thank you for the review. I agree with all of this. The two-layer split made sense when I wrote it, but you're right that the manual policy.check_all(&selection, &tx) step is invisible at the type level. Nothing prompts the caller to run it, and accepting both a Selection and a Transaction lets the two drift. I will fold the validation into construction and have create_psbt_with_policy build the tx it validates.

noahjoeris (commented):
Nice, thanks for updating 🔥

I’d wait now for a Concept ACK on this direction by @ValuedMammal before spending more time on details.

- One API point: I’d prefer to keep `create_psbt(...)`. It could either preserve the old unchecked behavior for compatibility, or become checked by default as `create_psbt(params, tip)` using `MempoolPolicy::default()`. I think we should discuss that explicitly. Right now most examples moved to `create_psbt_unchecked(...)`, which makes the unchecked path look like the normal path.
- Also, `MempoolPolicy` only models part of Core v30 policy for now, so we should expect it to grow as more checks are added.

ValuedMammal:
I did a first pass. Overall I like the idea and it appears to be well implemented. Biggest concern is the long term API stability as policies change and grow outdated.

Pros
- Conceptually I like the separation between standardness checks and post selection checks
- Certainly in favor of consensus-level validation to catch obvious mistakes in tx construction (e.g. negative fee)
- And maybe a subset of policy/standardness checks
- Important to differentiate tx building validation, bitcoin consensus, and local node policy
- From a maintenance point of view I would take a light touch approach to policy/standardness since it will likely be biased and incomplete


Cons
- `MempoolPolicy::default()` is going to be a maintenance liability since the defaults will inevitably change over time
- MempoolPolicy as a name is potentially misleading. We don't guarantee that if your tx passes MempoolPolicy, it will be accepted to mempools
- NegativeFee error is good for guarding against obviously invalid txs, but not exactly a policy preference

src/utils.rs
```diff
@@ -180,3 +180,17 @@ fn random_range(rng: &mut impl RngCore, n: u32) -> u32 {
...
+/// Excludes bare multisig (widely disabled at the relay layer).
+pub fn is_standard_script(script: &Script) -> bool {
+    script.is_p2pk()
+        || script.is_p2pkh()
+        || script.is_p2sh()
+        || script.is_p2wpkh()
+        || script.is_p2wsh()
+        || script.is_p2tr()
+        || script.is_op_return()
+}
```
ValuedMammal: We could collapse the segwit/taproot conditions to `script.is_witness_program()`, which should also cover pay-to-anchor?

src/selector.rs
```diff
@@ -320,8 +317,260 @@ impl SelectorParams {
...
+    pub fn build(self) -> Result<SelectorParams, SelectorParamsError> {
+        self.build_with_policy(&MempoolPolicy::default())
+    }
+
+    /// Validate against a caller-supplied policy.
+    ///
+    /// Output-side standardness checks (dust, `OP_RETURN`, standard script
+    /// types) are evaluated against `policy`. The dust feerate used by the
+    /// change-policy computation also defaults to `policy.dust_relay_feerate`
+    /// unless explicitly overridden via [`Self::dust_relay_feerate`].
```
ValuedMammal: Confused why the dust-feerate of the params would override the MempoolPolicy when we're supposed to be validating the specified params against a chosen policy.

aagbotemi:
> I did a first pass. Overall I like the idea and it appears to be well implemented. Biggest concern is the long term API stability as policies change and grow outdated.
> 
> Pros
> 
> * Conceptually I like the separation between standardness checks and post selection checks
> * Certainly in favor of consensus-level validation to catch obvious mistakes in tx construction (e.g. negative fee)
> * And maybe a subset of policy/standardness checks
> * Important to differentiate tx building validation, bitcoin consensus, and local node policy
> * From a maintenance point of view I would take a light touch approach to policy/standardness since it will likely be biased and incomplete
> 
> Cons
> 
> * `MempoolPolicy::default()` is going to be a maintenance liability since the defaults will inevitably change over time
> * MempoolPolicy as a name is potentially misleading. We don't guarantee that if your tx passes MempoolPolicy, it will be accepted to mempools
> * NegativeFee error is good for guarding against obviously invalid txs, but not exactly a policy preference

Thank you for the review. I overshot the scope. Re-reading the original TODO, this is where I got the misconception - "error on anything that doesn't satisfy mempool policy". I took that literally and tried to build toward it.

I can see that this implementation was tending toward tx broadcast preparation, which is outside the scope of `bdk-tx` as a tx construction library. Going forward I will narrow this PR to tx construction correctness. I'll remove `MempoolPolicy`, `ChainTip`, and the related post-selection policy functionality.

src/selector.rs
aagbotemi: This will be removed in the next push, along with MempoolPolicy itself.

evanlinjin:
> MempoolPolicy::default() is going to be a maintenance liability since the defaults will inevitably change over time.

Maybe instead of default, we just have constructors which construct MempoolPolicy of certain Bitcoin core versions

src/selection.rs
```diff
@@ -303,6 +329,24 @@ impl Selection {
...
+        let segwit_overhead = if self.inputs.iter().any(|i| i.is_segwit()) {
+            Weight::from_wu(2)
+        } else {
+            Weight::ZERO
+        };
+        let weight = psbt.unsigned_tx.weight() + satisfaction + segwit_overhead;
+
+        if weight > Weight::from_wu(MAX_STANDARD_TX_WEIGHT as u64) {
+            return Err(CreatePsbtError::MaxStandardTxWeightExceeded { weight });
+        }
```
noahjoeris: With this we don't have an unchecked path anymore for non standard txs right?

Also create_psbt is getting quite big. Should we add helper functions for the checks (weight, negativefee)?

src/selector.rs
```diff
@@ -320,8 +323,241 @@ impl SelectorParams {
...
+        }
+    }
+}
+
+/// Returns `true` if the given script is a recognized standard output script type.
+pub fn is_standard_script(script: &Script) -> bool {
+    script.is_p2pk()
+        || script.is_p2pkh()
+        || script.is_p2sh()
+        || script.is_witness_program()
```
noahjoeris: I'm not sure `script.is_witness_program()` is correct here. It allows scripts in the range of 2-40 bytes length. Which could be non-standard. 

Maybe it's better to add them directly.
```rust
script.is_p2wpkh()
    || script.is_p2wsh()
    || script.is_p2tr()
    || script.is_p2a()
```

src/selector.rs
```diff
@@ -320,8 +323,241 @@ impl SelectorParams {
...
+        || script.is_op_return()
```
noahjoeris: I think `is_op_return()` only checks the first byte which might not be enough.
e.g. `OP_RETURN OP_CHECKSIG` could pass the check

noahjoeris (commented):
Thanks for updating again! 🔥  I left some comments.

Could you outline where these checks come from and why we picked this subset? Are there other checks we are intentionally leaving out?

src/selector.rs
aagbotemi: Thanks for the catch. I've updated

src/selector.rs
aagbotemi: `script.is_p2a()` is not available yet, so i used a local helper

src/selection.rs
aagbotemi: We've only covered weight and output-side checks.
Helper functions have been added.

aagbotemi:
> Maybe instead of default, we just have constructors which construct MempoolPolicy of certain Bitcoin core versions

The whole `MempoolPolicy` struct, `ChainTip`, and the post-selection validation path are being removed. `bdk-tx` is a tx construction library, not a broadcast-preparation library and versioning Core's relay policy here is the wrong layer for that responsibility.

> Could you outline where these checks come from and why we picked this subset? Are there other checks we are intentionally leaving out?

After @ValuedMammal's review I've narrowed the scope significantly. The checks agreed upon will be construction invariants pre-selection, which are drawn from Bitcoin Core's `IsStandardTx`, narrowed to checks that hold regardless of which node or Core version you're targeting. Everything relay-readiness shaped (allowed_versions, locktime, min-relay feerate, witness stack limits, ChainTip-dependent checks) is out of scope for this library as they are transaction preparedness library.

src/selection.rs
```diff
@@ -106,6 +109,13 @@ pub enum CreatePsbtError {
     OutputUpdate(miniscript::psbt::OutputUpdateError),
     /// Occurs when applying anti-fee-sniping fails.
     AntiFeeSniping(AntiFeeSnipingError),
+    /// Total output value exceeds total input value.
+    NegativeFee,
```
evanlinjin: Is this even possible?

evanlinjin (changes requested):
Approach NACK

I'd like to push back on the overall direction here. Mempool policy is being baked into `SelectorParams` (and into module-level constants), but mempool policy is not a constant — it varies by Bitcoin Core version (P2A became standard in v28, OP_RETURN policy changed materially in v30), by node implementation (Knots, btcd), and by node operator tuning (`-dustrelayfee`, `-datacarriersize`, `-maxmempool`). Encoding one frozen-in-time snapshot inside the selector hides that variation and makes it untunable.

Concretely, the current PR scatters policy in several places:

- `dust_relay_feerate` is a field on `SelectorParams`.
- `MAX_OP_RETURN_BYTES` is a `pub const` in `selector.rs`.
- `MAX_STANDARD_TX_WEIGHT` is hardcoded inside `Selection::check_max_standard_weight`.
- The standard-script allowlist is hardcoded in `is_standard_script`.

These are all the same kind of thing, presented as if there's one true answer.

I'd prefer a dedicated `StandardnessPolicy` type that owns this surface:

```rust
#[non_exhaustive]
pub struct StandardnessPolicy {
    pub dust_relay_feerate: FeeRate,
    pub max_standard_tx_weight: Weight,
    pub max_op_return_bytes: usize,
    // ...standard output kinds, datacarrier flags, etc.
}

impl StandardnessPolicy {
    pub fn bitcoin_core_v28() -> Self { ... }
    pub fn bitcoin_core_v29() -> Self { ... }
    pub fn bitcoin_core_v30() -> Self { ... }

    pub fn check_output(&self, _: &Output)     -> Result<(), _>;
    pub fn check_input(&self,  _: &Input)      -> Result<(), _>;
    pub fn check_tx(&self,     _: &TxTemplate) -> Result<(), _>;
}
```

The name maps 1:1 to Bitcoin Core's `IsStandardTx` — these are exactly the rules that function checks. `RelayPolicy` or `MempoolPolicy` would be fuzzier, since "relay" / "mempool" also cover stateful concerns (min-relay-fee, RBF rules, ancestor limits) that explicitly do not belong here (see point 4 below).

`SelectorParams` would drop `dust_relay_feerate` and the standardness checks. Callers can validate outputs pre-selection (`policy.check_output`), the full shape post-template (`policy.check_tx`), or skip validation entirely if they're deliberately producing non-standard txs. The selector itself stays policy-agnostic.

A few specific things worth pinning down before any rework:

1. **Change-output dust threshold.** `to_cs_change_policy` currently consumes `dust_relay_feerate` directly. If policy moves out, either pass `&StandardnessPolicy` at call time or keep just the feerate field on `SelectorParams`. I'd lean toward passing the policy — keeps `SelectorParams` clean.

2. **`check_input` scope.** The standard *prevout* script type is already covered by checking outputs we constructed. The remaining input-side rules are mostly satisfaction-size limits (scriptSig ≤ 1650 B, P2WSH script ≤ 3600 B, taproot stack item sizes). Worth being concrete in the design — otherwise it's a method that's hard to specify.

3. **Block on #73.** `check_tx(&TxTemplate)` is the right surface but `TxTemplate` doesn't exist on `master` yet. Land #73 first; that also gives a natural place for a fluent `tx_template.check_against(&policy)` step.

4. **Scope ceiling.** Be explicit that this models *static standardness*, not full mempool acceptance. Ancestor/descendant counts, RBF rules 2/3/4, and mempool-min-fee are stateful and don't belong here — the name `StandardnessPolicy` makes that boundary self-evident.

One note on the existing `Selection::create_psbt` checks: `check_max_standard_weight` is pure policy and should move into `StandardnessPolicy::check_tx`. `check_negative_fee` is a correctness invariant rather than policy — that one can stay where it is.

aagbotemi:
@evanlinjin Thanks for the review. I want to push back on the architectural direction, though I agree with some of the observations.

bdk-tx is a transaction *construction* library. The checks in this PR are deliberately minimal construction-bug guards. The four locations you flagged (`dust_relay_feerate`, `MAX_OP_RETURN_BYTES`, `MAX_STANDARD_TX_WEIGHT`, `is_standard_script`) are a fixed floor of sanity checks the library always enforces and not a partial mempool policy.

Following earlier discussion with @ValuedMammal , I think it's okay for configurable standardness/relay validation to belong in a separate library scoped to that concern. For reference, this is the [design notes](https://hackmd.io/@aagbotemi/BJUc7_yyzl), a tx *preparedness* library will have a fundamentally different maintenance contract like tracking Core releases, node implementations, which are exactly the burden you identified.

Earlier reviews on this PR pushed back on `MempoolPolicy::default()` and the version-tracking burden it implied. The proposed `StandardnessPolicy::bitcoin_core_v28/v29/v30()` constructors relocate that burden but don't eliminate it because every future Core release that changes relay rules still forces a decision in bdk-tx.

Once `StandardnessPolicy` exists, every future addition (check_input, min_relay_feerate, ancestor/descendant accounting) may become "we already have a policy type, why not add this too?", which is exactly why the abstraction belongs in bdk_mempool, not bdk-tx.

**Some of the points you raised:**
1. **Change-output dust threshold** Stays on SelectorParams because `to_cs_change_policy` needs it to compute change min_value.
2. **`check_input` scope.** Agreed it's hard to specify cleanly. Out of scope here, belongs in bdk_mempool if anywhere.
3. **Scope ceiling.** Agreed. I'll add a module-level doc stating bdk-tx does output-side construction-bug checks only.

**What I suggest we change**
- **Move `check_max_standard_weight` to opt-in.** You're right that this is policy, not invariant. I'll move it to a separate `Selection::check_standardness` method called explicitly.
- **`check_negative_fee` stays where it is.** Construction invariant, not policy.
- **Tighten doc comments** on `check_standardness` and the weight check to state what they are not doing.

evanlinjin:
@aagbotemi have you curated this AI generated response before posting it?

Standardness policy by definition is a subset of mempool policy.

The response is self contradictory. I'm not even sure how to respond to it.

src/selection.rs
evanlinjin: `check_negative_fee` is unreachable since the coin selector already errors when there is insufficient funds.

This is essentially an internal logic check exposed as an external API. We should instead just have an `assert` statement.

evanlinjin:
Looking at the `bdk_mempool` plan, I propose splitting the logic up into the following crates:

* `bdk_standardness` - pure [`IsStandardTx`](https://github.com/bitcoin/bitcoin/blob/00af5620f01dbd1d2e696487303fad0382b67591/src/policy/policy.cpp#L100).
* `bdk_mempool` - has awareness of mempool snapshot, total UTXO set to determine whether the candidate tx will be accepted. Depends on `bdk_standardness`.
* `bdk_tx` - will use `bdk_standardness` for sanity checks.

I think it makes sense to turn this repo into a cargo workspace to contain the above. cc. @ValuedMammal @nymius 

In regards to what remains in this PR (if we are going forward with this proposal):

* `check_negative_fee` is unreachable and should be removed.
* Everything else is `IsStandardTx` checks and should be implemented in `bdk_standardness`.

aagbotemi:
> @aagbotemi have you curated this AI generated response before posting it?

@evanlinjin the summary of the response I gave is trying to explain that the bdk-tx is a transaction building library and trying to stay in that scope. I understand that Standardness Policy is a subset of Mempool Policy, but only pointing out maintenance burden of having v28/v29/v30/v31 and so on, for a tx building library.

> I think it makes sense to turn this repo into a cargo workspace to contain the above.

I think your proposal to split into a cargo workspace makes sense, such that the logic will be shared into different crates.

ValuedMammal:
I agree the Bitcoin Core version specific policies are overkill for right now, and we should instead focus on structural validity. Not sure what that leaves to be implemented since we've ruled out the negative fee possibility. While some of these technically fall under standardness I'm not against the idea of including basic sanity checks that we define and aren't likely to change very often. Of course these are guardrails and not hard and fast rules.

- is standard spk
- min/max tx weight
- output below dust
- max feerate
- max burn amount

evanlinjin:
> I agree the Bitcoin Core version specific policies are overkill for right now

What do you mean by overkill?

> we should instead focus on structural validity. Not sure what that leaves to be implemented since we've ruled out the negative fee possibility.

Yes, `bdk_tx` should never create structurally invalid transactions that can never enter a block.

> While some of these technically fall under standardness I'm not against the idea of including basic sanity checks that we define and aren't likely to change very often. Of course these are guardrails and not hard and fast rules.

There are two concerns here. One is `IsStandardTx`, the other is UX related.

- **Standardness:** Is standard spk?, min/max tx weight, output below dust, etc.
- **UX:** Max feerate, Max burn amount, Fee/send-amount ratio (maybe).

Both of these concerns can be categorized under "basic sanity checks". However, I think it makes sense for the UX checks to exist in this crate (if we decide to have them). The values are ours to define, they don't change with Core releases.

---

Let me clarify my reasoning for the proposed direction.

### Why did I suggest a separate `bdk_standardness` crate?

Both `bdk_mempool` and `bdk_tx` (with basic sanity checks) need standardness checks. It's best to implement these somewhere accessible to both.

If we implement standardness checks in `bdk_tx`, we either need to make `bdk_mempool` depend on `bdk_tx`, or duplicate those checks in `bdk_mempool`. Both are non-ideal:

- **Wrong dependency direction.** A mempool checker importing a tx-construction lib is upside down.
- **Duplication drifts.** Two implementations of the same policy guarantee that whichever is wrong at any given moment is a bug waiting to bite a specific caller.

`bdk_tx` would `pub use bdk_standardness as standardness;`, so user-facing call sites are identical to having the checks in `bdk_tx` directly. The split is for the dependency DAG, not for users.

The other option is to have the standardness checks in `bdk_mempool` and `bdk_tx` will just depend on that. However, that might be worse than having `bdk_standardness` because that pulls in the UTXO & mempool machinery that the majority of callers will not need.

### Why `StandardnessPolicy` as a struct?

- **Standardness policy is not constant — changes happen relatively often and different node implementations have different policies.** Empirically, Bitcoin Core ships standardness changes ~2-4 times per year, and essentially every major release touches something. Recent examples that matter for tx construction:
  - **v28.0** (Oct 2024): P2A (BIP433) made standard, TRUC v3 transactions made standard, full-RBF made the default.
  - **v30.0** (Oct 2025): `datacarriersize` default raised 83 → **100,000** bytes; multiple OP_RETURNs per tx allowed; per-tx legacy sigop cap added.

  Concretely, the PR-as-written hardcodes `MAX_OP_RETURN_BYTES = 100_000`. That value is the v30.0 default; the prior default was 83 bytes. The PR is already implicitly tracking the v30.0 default without naming the version. Anyone depending on `bdk_tx` today gets v30.0 OP_RETURN semantics whether they intended to or not. A `StandardnessPolicy` struct makes that choice explicit (or at minimum, introspectable).

  Across node implementations the divergence is larger: Bitcoin Knots defaults `datacarriersize` to **42 bytes** (vs Core's 100,000) and applies stricter relay filtering generally; Libre Relay deliberately relays consensus-valid-but-non-standard transactions; alternative implementations (btcd, etc.) carry their own policy code. A struct lets callers target a specific deployment instead of `bdk_tx`'s choice.

- **Separation of concern.** Having a separate `StandardnessPolicy` struct tells us explicitly what it is without relying on the docs. It's also introspectable at runtime — "what policy was applied" becomes a value you can inspect, log, or compare against the actual relaying node's policy when debugging a rejected broadcast, instead of a source-code question.

- **Rare, but it's possible that not everybody wants to create a standard transaction. `StandardnessPolicy` allows the caller to define its own.** Real cases: protocol R&D, high-data-carrier applications, miner direct-to-block, CPFP sponsors in package relay. With a `StandardnessPolicy` they pass a relaxed or custom-tuned policy and still get `bdk_tx`'s construction help. Without it, the only escape hatch is `build_unchecked` — all-or-nothing.

src/selector.rs
nymius: It seems you have a legitimate use case for `is_p2a` in `rust-bitcoin@0.32.x`. I recommend you to backport this PR https://github.com/rust-bitcoin/rust-bitcoin/pull/5199 then.

nymius:
> Looking at the `bdk_mempool` plan, I propose splitting the logic up into the following crates:
> 
>     * `bdk_standardness` - pure [`IsStandardTx`](https://github.com/bitcoin/bitcoin/blob/00af5620f01dbd1d2e696487303fad0382b67591/src/policy/policy.cpp#L100).
> 
>     * `bdk_mempool` - has awareness of mempool snapshot, total UTXO set to determine whether the candidate tx will be accepted. Depends on `bdk_standardness`.
> 
>     * `bdk_tx` - will use `bdk_standardness` for sanity checks.
> 
> 
> I think it makes sense to turn this repo into a cargo workspace to contain the above. cc. @ValuedMammal @nymius
> 
> In regards to what remains in this PR (if we are going forward with this proposal):
> 
>     * `check_negative_fee` is unreachable and should be removed.
> 
>     * Everything else is `IsStandardTx` checks and should be implemented in `bdk_standardness`.

Sorry, I'm lost here, when `bdk_mempool` was discussed? or `bdk_standardness`? It is something that just started here?

@aagbotemi I need more specificity with this question `What would a miner do with this mempool?`, when you say mempool, what's the representation of it that you're using? Why originally it was related to `bdk-tx`? Also, is this proposal something you just created for this PR or it's something you will be working on for more than a single PR?

If we want to have `standarness` validation here, I wouldn't start with a workspace unless it becomes necessary. It could still have the semantic separation if it is placed in a separated module. Let's first decide what we put inside. Also, we don't have a major version yet, so it shouldn't be an API concern now.

> `bdk_tx` should never create structurally invalid transactions that can never enter a block.

I would optimize this library to just do that then.

aagbotemi:
> Sorry, I'm lost here, when bdk_mempool was discussed? or bdk_standardness? It is something that just started here? Why originally it was related to bdk-tx?

@nymius It started here in this PR. `SelectorParams` has a TODO that one of its tasks should `error on anything that doesn't satisfy mempool policy`. That made me have somewhat an elaborate design. Upon review, the point was raised that bdk-tx is a `tx construction` library and the PR at that point was starting to pull `tx broadcast preparation` into it. The conclusion was that bdk-tx should keep only lightweight sanity checks, and the heavier pieces,​ `relay-readiness`,  `policy validation` and cluster mempool implementation should live in a seperate crate(bdk-mempool) rather than here. That keeps the mempool-policy maintenance burden off bdk-tx.

> I need more specificity with this question `What would a miner do with this mempool?`, when you say mempool, what's the representation of it that you're using

The mempool is represented as a set of clusters, where each cluster is a connected component of the unconfirmed dependency graph. With `bdk_mempool`, the miner knows the block template, which is the ordered set of txs the miner should include in the next block to maximize the fee revenue, each cluster is linearized into a topological order, then chunked into groups, and the chunks are selected by highest feerate first until the block is full.

> is this proposal something you just created for this PR or it's something you will be working on for more than a single PR?

It's more than a single PR. This proposal though came up through this PR, but will be in a seperate crate, rather than in bdk-tx, since bdk-tx is a tx construction library.

nymius:
`error on anything that doesn't satisfy mempool policy` is a bit stricter than `bdk_tx should never create structurally invalid transactions that can never enter a block.`, which one should `bdk-tx` aim for? I agree this is a reason to separate concerns, but the elaboration of the problem is getting more complex than what I initially thought.

Was something like the proposed solution what triggered the author to create the **TODO**?
As a library I wouldn't provide consistency with a policy as a node does, so I wouldn't precipitate to create a `bdk_mempool` yet.

src/selector.rs
aagbotemi: Thank you for pointing me to that

evanlinjin:
> which one should bdk-tx aim for? 

@nymius I think `bdk_tx` should only focus on `IsStandardTx` as a sane default. The `bdk_mempool` idea (where there is full UTXO-set and mempool awareness) also needs `IsStandardTx`.

That's why I suggest, long term, to extract transaction standardness logic (`IsStandardTx`) into a separate crate.

For now though, I'm happy with keeping standardness logic in this crate as a submodule for now to get the API shape correct and figure out how it can be integrated in our `bdk_tx` stages.

> the elaboration of the problem is getting more complex than what I initially thought.

We basically want `IsStandardTx` checks and what constitutes a standard tx is evolving, there are multiple versions of it, and some callers want custom definitions.

This is my complete understanding of the problem (if it helps).

> Was something like the proposed solution what triggered the author to create the TODO?

Where's the TODO?

src/selection.rs
```diff
@@ -106,6 +107,11 @@ pub enum CreatePsbtError {
     OutputUpdate(miniscript::psbt::OutputUpdateError),
     /// Occurs when applying anti-fee-sniping fails.
     AntiFeeSniping(AntiFeeSnipingError),
+    /// Transaction weight exceeds the consensus block-weight limit.
+    ExceedsBlockWeight {
+        /// The estimated weight of the transaction.
+        weight: Weight,
+    },
```
evanlinjin: This should not be part of the PSBT-level concern. Technically, neither is `AntiFeeSniping` (and that is being moved out in #73).

We need to somehow integrate this check in the coin selection stage.

src/selection.rs
```diff
@@ -329,6 +342,15 @@ impl Selection {
             }
         }
 
+        // Guard against exceeding `MAX_BLOCK_WEIGHT`.
+        let weight = self
+            .max_signed_weight(&psbt.unsigned_tx)
+            .unwrap_or(Weight::MAX);
+
+        if weight > Weight::from_wu(MAX_BLOCK_WEIGHT as u64) {
```
evanlinjin: This should not be a hardcoded value. I think it makes sense to be a parameter in `SelectorParams`. Then this value can be set to a certain standardness-policy version (maybe in a different PR though).

```rust
let params = SelectorParams::new(feerate, outputs, change)
    .with_standardness(StandardnessPolicy::bitcoin_core_v30());
```

src/selector.rs
```diff
@@ -329,6 +326,113 @@ impl SelectorParams {
...
         )
     }
+
+    /// Start a builder.
+    ///
+    /// The two required fields are taken eagerly so the builder cannot be
+    /// constructed in an incomplete state. Outputs and optional fields are
+    /// added with chained setters; the builder terminates with
+    /// [`build`](SelectorParamsBuilder::build).
+    pub fn builder(target_feerate: FeeRate, change_script: ChangeScript) -> SelectorParamsBuilder {
```
evanlinjin: This builder pattern is not adding anything new, I would remove it for now.

I'm assuming the idea is that the standardness policy will be contained within the builder so that methods will have checks?

I.e.

```rust
// Builder is constructed with `StandardnessPolicy`.
// Returns error if standardness policy is not satisfied.
builder.push_output(output)?
```

evanlinjin (commented):
I think we should focus this PR on just adding a `SelectorParams::max_weight` parameter.

aagbotemi:
Thank you for the review @evanlinjin, I’ll work on the fixes

nymius:
> That's why I suggest, long term, to extract transaction standardness logic (`IsStandardTx`) into a separate crate.
> 
> For now though, I'm happy with keeping standardness logic in this crate as a submodule for now to get the API shape correct and figure out how it can be integrated in our `bdk_tx` stages.

Fair enough, this aligns with the structural validity you were looking for @ValuedMammal?

> > Was something like the proposed solution what triggered the author to create the TODO?
> 
> Where's the TODO?

It was mention somewhere in the comments that the reason this PR was created was because of the existence of a TODO in the `SelectorParams`, @aagbotemi , could you link that here?

aagbotemi:
> > Where's the TODO?
> 
> TODO in the SelectorParams, @aagbotemi , could you link that here?

The TODO is linked here https://github.com/bitcoindevkit/bdk-tx/blob/b331ab3a80ba62a767fc56ddf22437fb0775f020/src/selector.rs#L26-L31

evanlinjin (commented):
Right now `max_weight` is checked after a selection is complete. Whereas `max_weight` should really steer selection since the goal is to create a valid tx.

To do this properly, we need upstream changes in `bdk_coin_select`: `MaxWeight<M>` metric that wraps metric `M` which changes the `score` impl to return `None` for selections that overshoots a max-weight upper bound.

aagbotemi:
Thank you for the review @evanlinjin. I will make the upstream change to `bdk_coin_select`

evanlinjin:
@aagbotemi I'm actually not sure whether `MaxWeight<M>` is better or `(M, MaxWeight)` where the metric is meant to be used adjacently is better. I'm assuming `MaxWeight<M>` is.

evanlinjin:
@aagbotemi Let me know when the PR is ready to review and please explain the design decisions so it's easier to review.

I'll convert this to a draft before then.

aagbotemi:
Thanks @evanlinjin. This is ready for review now, I've added a "Notes for reviewers" section to the description explaining the design decisions.
One thing to note: this PR depends on [bdk_coin_select#48](https://github.com/bitcoindevkit/coin-select/pull/48) (Target::max_weight). The Cargo.toml currently points at a temporary git dependency so the two can be reviewed together. Once it is merged and released, I'll switch it back to the published version. So it's open for review now, but can't be merged until it lands.

evanlinjin:
@aagbotemi what I meant by explain is to explain why things are done the way they are. I did not mean describe what you did - that is inherently obvious by looking at the code.

For example, why the error types were changed the way they are.

PR #50, opened 2026-04-25
fix(rbf)!: include evicted descendants in replacement fee floor

noahjoeris:
### Description
**Problem:** When replacing a tx that has unconfirmed descendants in the mempool, those descendants are evicted along with it but their fees were not counted toward the replacement's required fee. This allowed under-paying replacements that a node would reject.

**Fix:**
- `extract_replacements` now sums descendant fees and passes the total to `RbfSet::new` / `RbfParams::new`.
- `RbfSet::new` gains a `descendant_fee: Amount` parameter (breaking change).
- `RbfParams` gains a `descendant_fee: Amount` field. `to_cs_replace()` includes it in the absolute fee floor; `max_feerate()` continues to look at direct conflicts only.

### DIFF
--- a/src/rbf.rs
+++ b/src/rbf.rs
@@ -9,7 +9,11 @@ use crate::{CanonicalUnspents, Input, RbfParams};
 
 /// Set of txs to replace.
 pub struct RbfSet {
+    /// Original txs the replacement must conflict with.
     txs: HashMap<Txid, Arc<Transaction>>,
+    /// Sum of fees from evicted descendants, counted toward the replacement fee floor.
+    descendant_fee: Amount,
+    /// Previous outputs spent by the originals.
     prev_txouts: HashMap<OutPoint, TxOut>,
 }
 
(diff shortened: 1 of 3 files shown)


evanlinjin:
What's the rationale for having a separate field for descendants of the originals?

What if we just document that if you forget to include descendants in `RbfSet::txs`, you'll get an inaccurate result?

src/rbf.rs
```diff
@@ -152,6 +172,11 @@ impl RbfSet {
 
     /// Coin selector RBF parameters.
     pub fn selector_rbf_params(&self) -> RbfParams {
-        RbfParams::new(self.txs.values().map(|tx| (tx.as_ref(), self._fee(tx))))
+        RbfParams::new(
```
evanlinjin, line 162: `RbfParams` should have an additional field `RbfParams::descendant_fee`.

`RbfParams::max_feerate` should return the highest feerate of one original (directly conflicting) transaction. This PR currently also looks at descendants of the original transactions which is wrong.

Note that `RbfParams::to_cs_replace` should still return a `Replace::fee` inclusive of descendants (provided by `RbfParams::descendatn_fee`.

evanlinjin (changes requested):
Thanks for catching this bug.

noahjoeris:
Thanks for the review!! I addressed the issue.

evanlinjin (approved):
ACK 5c49d46cf91fb73d4795662a8165217f51581784

evanlinjin:
@noahjoeris Sorry, needs a rebase!

evanlinjin:
@noahjoeris Code wise, it's looking really good. A few things need fixing:

* The PR description is stale since your last change - please give it an update.

* Commit message can be a bit more descriptive and forgot to signal API breakage.

  **Suggestion:**

  ```
  fix(rbf)!: include evicted descendants in replacement fee floor

  When an original tx has unconfirmed mempool descendants, RBF evicts the
  descendants along with it so descendant fees must be covered by the
  replacement. Previously, RbfParams::to_cs_replace().fee` summed the
  originals without the descendants.

  `extract_replacements` now sums descendant fees during the eviction
  walk and passes the total to `RbfSet::new`/`RbfParams::new`.
  ```

noahjoeris:
I updated, thanks!
I'm still a bit new to some of those things in open source. I will pay more attention to those details in the future. Appreciate the patience 💙

evanlinjin:
@noahjoeris No worries, keep up the good work.

evanlinjin (approved):
ACK 007d0edf84cd37885b2f50cad20c3927aefbce87

PR #51, opened 2026-04-25
fix(deps): replace deprecated FeeRate::from_sat_per_vb_unchecked

noahjoeris:
### Description
- `FeeRate::from_sat_per_vb_unchecked` got deprecated in `bitcoin-units 0.1.3`, which broke CI
- I swapped it for `FeeRate::from_sat_per_vb(n)` everywhere
- I went with `from_sat_per_vb` instead of `from_sat_per_vb_u32` for compatibility. `from_sat_per_vb_u32` would require a direct bitcoin-units >= 0.1.3 constraint; bitcoin only requires bitcoin-units = "0.1.0", so it can still resolve to versions without that API


ValuedMammal (approved):
ACK a6217bf97bb7003e073254dc89f50282da614b0e

PR #65, opened 2026-05-12
fix(afs)!: preserve input timelock requirements

### LINKED ISSUE #62: AFS locktime-branch may undershoot CLTV requirements of an input
### Description

In `create_psbt`, the `current_height` provided to `apply_anti_fee_sniping` is the result of `accumulate_max_locktime`. In other words, the `current_height` value might come from a CLTV requirement of an input, and be the minimum required locktime for the transaction to be consensus valid.

In the locktime branch of `apply_anti_fee_sniping`, there is a 10% change of reducing this locktime (`tx.locktime = current_height - r` where `0 <= r <= 100`). Therefore, `tx.locktime` might be set to a value that makes a transaction consensus invalid.

### Severity

~10% of AFS-locktime-branch txs has an offset applied. If there is an CLTV input whose min locktime is close to the `current_height` value provided for AFS, `create_psbt` may create a consensus-invalid transaction.

### Proposed Solution

* Rename `PsbtParams::fallback_locktime` to `min_locktime`. This is a better name as it describes a floor to the locktime and that it's a contributor to `accumulate_max_locktime`.

* Rename the `current_height` input of `apply_anti_fee_sniping()` to `tip_height`. This signals that it's the tip height.

* Call `accumulate_max_locktime` in `apply_anti_fee_sniping()` to get `acc_locktime`. `tx.locktime` should be set to `max(acc_locktime, tip_height)` where `tip_height` has a 10% chance of an offset.

This way, CLTV is always respected and `min_locktime` is always enforced.

### LINKED ISSUE #63: AFS sequence branch sets `tx.locktime` to `LockTime::ZERO` unconditionally
### Description

Like the bug described in #62, forcing `tx.locktime` to ZERO may result in creating a consensus-invalid transaction as inputs may contain CLTV requirements.

### Severity

May create consensus-invalid transactions.

### Proposed Solution

If the `accumulate_max_locktime` value is non-zero, always use the AFS-locktime-branch.

### LINKED ISSUE #64: In the sequence-branch of AFS, we may force a sequence value that is incompatible with CSV requirements
### Description

Some Taproot inputs may be CSV (relative timelock) requirements. Changes made to `tx.input[i].sequence` must still satisfy CSV. The current code base does not guarantee this.

### Severity

Can result in consensus-invalid transactions.

### Proposed Solution

Inputs with CSV requirements should be filtered out for the `taproot_inputs` pool.

```rust
input.relative_timelock().is_none()
```

aagbotemi:
### Description

Fixes consensus-validity issues in BIP-326 anti-fee-sniping.

Closes #62
Closes #63
Closes #64

### Bugs

- The **nLockTime strategy** could overwrite `tx.lock_time` with a value lower than the accumulated input CLTV.
- The **nSequence strategy** zeroed `tx.lock_time`, also breaking accumulated CLTV.
- The **nSequence strategy** could select a Taproot input that already had a relative-timelock (CSV) requirement and overwrite its sequence.

### Fixes

- The locktime path only updates `tx.lock_time` when the AFS-proposed value still satisfies the existing requirement.
- The sequence path leaves `tx.lock_time` alone and filters CSV-bearing Taproot inputs out of the candidate pool.

### Breaking changes

- `PsbtParams::enable_anti_fee_sniping: bool` + `fallback_locktime` → `anti_fee_sniping: Option<absolute::Height>` + `min_locktime`. The tip height is now required when AFS is enabled, and time-based tips are unrepresentable.
- `apply_anti_fee_sniping` is now `pub(crate)`; `rbf_enabled` param removed (derived from `tx.is_explicitly_rbf()`).
- New `AntiFeeSnipingError` type, wrapped under `CreatePsbtError::AntiFeeSniping`.

### DIFF
--- a/examples/anti_fee_sniping.rs
+++ b/examples/anti_fee_sniping.rs
@@ -88,13 +88,10 @@ fn main() -> anyhow::Result<()> {
                 },
             )?;
 
-        let fallback_locktime: LockTime = LockTime::from_consensus(tip_height.to_consensus_u32());
-
         let selection_inputs = selection.inputs.clone();
 
         let psbt = selection.create_psbt(PsbtParams {
-            enable_anti_fee_sniping: true,
-            fallback_locktime,
+            anti_fee_sniping: Some(tip_height),
             ..Default::default()
         })?;
 
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -285,23 +285,24 @@ mod tests {
     use super::*;
     use bitcoin::{
         absolute::{self, LockTime, Time},
+        relative,
         secp256k1::Secp256k1,
         transaction::{self, Version},
-        Amount, ScriptBuf, Transaction, TxIn, TxOut,
+        Amount, ScriptBuf, Sequence, Transaction, TxIn, TxOut,
     };
     use miniscript::{plan::Assets, Descriptor, DescriptorPublicKey};
     use rand_core::OsRng;
 
     const TEST_DESCRIPTOR: &str = "tr([83737d5e/86h/1h/0h]tpubDDR5GgtoxS8fJyjjvdahN4VzV5DV6jtbcyvVXhEKq2XtpxjxBXmxH3r8QrNbQqHg4bJM1EGkxi7Pjfkgnui9jQWqS7kxHvX6rhUeriLDKxz/0/*)";
     const TEST_DESCRIPTOR_PK: &str = "[83737d5e/86h/1h/0h]tpubDDR5GgtoxS8fJyjjvdahN4VzV5DV6jtbcyvVXhEKq2XtpxjxBXmxH3r8QrNbQqHg4bJM1EGkxi7Pjfkgnui9jQWqS7kxHvX6rhUeriLDKxz/0/*";
+    const TEST_HEX_PK: &str = "032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3";
 
     fn setup_cltv_input(
         cltv: absolute::LockTime,
     ) -> anyhow::Result<(Input, Descriptor<DescriptorPublicKey>)> {
         let secp = Secp256k1::new();
-        let pk = "032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3";
-        let desc_str = format!("wsh(and_v(v:pk({pk}),after({cltv})))");
-        let desc_pk: DescriptorPublicKey = pk.parse()?;
+        let desc_str = format!("wsh(and_v(v:pk({TEST_HEX_PK}),after({cltv})))");
+        let desc_pk: DescriptorPublicKey = TEST_HEX_PK.parse()?;
         let (desc, _) = Descriptor::parse_descriptor(&secp, &desc_str)?;
         let plan = desc
             .at_derivation_index(0)?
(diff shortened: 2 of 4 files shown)


src/utils.rs
```diff
@@ -64,20 +83,22 @@ use rand_core::RngCore;
 pub fn apply_anti_fee_sniping(
     tx: &mut Transaction,
     inputs: &[Input],
-    current_height: absolute::Height,
-    rbf_enabled: bool,
+    tip_height: absolute::Height,
+    acc_locktime: absolute::LockTime,
```
evanlinjin: If this function is intended to be publicly exposed (which I assume it is due to `pub` and expansive doccomments), then we shouldn't have inputs that can conflict (`acc_locktime` is derived from `inputs` so we shouldn't have both). Recalculating `acc_locktime` internally is not expensive.

Since `acc_locktime` is calculated from `inputs` and `min_locktime`, we should just have `min_locktime` as an additional input.

src/selection.rs
```diff
@@ -245,16 +198,8 @@ impl Selection {
...
 
-        if params.enable_anti_fee_sniping {
-            let rbf_enabled = tx.is_explicitly_rbf();
-            let current_height = match tx.lock_time {
-                LockTime::Blocks(height) => height,
-                LockTime::Seconds(_) => {
-                    return Err(CreatePsbtError::InvalidLockTime(tx.lock_time));
-                }
-            };
-
```
evanlinjin, line 250: I understand that I suggested that we remove `InvalidLockTime` in the post-merge review of #5. However, to do it properly, we need to have more sophisticated logic (i.e. only allow the sequence branch when we have time-based locktime). I suggest that we introduce this in a separate PR.

For this PR, let's reintroduce this check and add an error variant; `AntiFeeSnipingError::UnsupportedLockTime`.

src/utils.rs
```diff
@@ -87,45 +108,52 @@ pub fn apply_anti_fee_sniping(
...
+    // Conditions that force nLockTime (vs nSequence).
     let must_use_locktime = inputs.iter().any(|input| {
-        let confirmation = input.confirmations(current_height);
+        let confirmation = input.confirmations(tip_height);
         confirmation == 0
             || confirmation > MAX_RELATIVE_HEIGHT
             || !input.prev_txout().script_pubkey.is_p2tr()
     });
 
+    let acc_locktime_is_nonzero = acc_locktime.to_consensus_u32() != 0;
```
evanlinjin: We can remove this as in line 137, you've provided an alternative fix for #63

evanlinjin (changes requested):
Thanks for working on this.

PR description is misleading - none of these bugs were flagged in the post-merge review. It also says this "fixes two consensus-validity issues", however this PR attempts to close 3 issues?

src/utils.rs
evanlinjin, line 1: Can we rename this module to `afs.rs`?

evanlinjin (changes requested):
There are two versions of the `random_probability` and `random_range` methods. The std versions add no meaningful value imo - I would just remove them.

src/utils.rs
evanlinjin: Moved to separate issue as this is bikeshedding.

aagbotemi:
Thank you for the review

evanlinjin (approved):
ACK 8487ab4e2aacd08022892a5bf2fc7abfdca611d0

src/utils.rs
```diff
@@ -1,20 +1,54 @@
...
+#[derive(Debug, Clone)]
+pub enum AntiFeeSnipingError {
+    /// Transaction `version` must be >= 2 for AFS to use relative locktimes.
+    UnsupportedVersion(Version),
+    /// AFS only supports height-based locktimes. A time-based (MTP) locktime
+    /// — either from `min_locktime` or from an input's CLTV requirement —
+    /// cannot be combined with the height semantics used here.
+    UnsupportedLockTime(absolute::LockTime),
+    /// Inputs have conflicting locktime units.
+    MixedInputLockTimeUnits,
```
evanlinjin: It's unfortunate we have a duplicate error variant - `MixedInputLokcTimeUnits` is checking the same thing as `PsbtError::LockTypeMismatch`.

Sorry for going back and forth, but how about we make `anti_fee_sniping` private and have `acc_locktime` as an input again?

evanlinjin (changes requested):

src/utils.rs
aagbotemi: Done in 3b72e6dd753267fd27b621ae658ceea58769b75d

src/selection.rs
```diff
@@ -53,67 +50,22 @@ pub struct PsbtParams {
...
-    /// certain conditions forcing nLockTime usage (unconfirmed inputs, non-Taproot inputs,
-    /// RBF disabled, etc.).
-    ///
-    /// # Error Cases
-    /// - Returns [`CreatePsbtError::InvalidLockTime`] if the locktime is not a block height
-    /// - Returns [`CreatePsbtError::UnsupportedVersion`] if transaction version is less than 2
-    ///
-    /// # Default
-    /// - Disabled by default (`false`).
-    ///
```
evanlinjin, line 77: Is there a reason why all of these docs are removed? This is the only public entry to AFS.

It definitely needs an update from our changes though.
Examples is no longer needed (I agree) since the API is less footgun-ey.

aagbotemi: I have restored the docs

src/utils.rs
```diff
@@ -64,20 +89,22 @@ use rand_core::RngCore;
 pub fn apply_anti_fee_sniping(
```
evanlinjin:
```suggestion
pub(crate) fn apply_anti_fee_sniping(
```

src/utils.rs
```diff
@@ -27,35 +55,32 @@ use rand_core::RngCore;
...
 ///
 /// # Errors
-/// Returns an error if:
-/// - Transaction version is less than 2 [`CreatePsbtError::UnsupportedVersion`]
+/// [`AntiFeeSnipingError::UnsupportedVersion`] if `tx.version < 2`.
 ///
 /// # Example
 /// ```ignore
-/// # use bdk_tx::Input;
+/// # use bdk_tx::{apply_anti_fee_sniping, Input};
```
evanlinjin: Remove this example. `use bdk_tx::apply_anti_fee_sniping` won't compile if we make this private.

src/utils.rs
```diff
@@ -27,35 +55,32 @@ use rand_core::RngCore;
 /// # Parameters
 /// - `tx`: The transaction to modify
 /// - `inputs`: The inputs associated with the transaction
-/// - `current_height`: The current blockchain height (used as the base for time locks)
-/// - `rbf_enabled`: Whether Replace-By-Fee is enabled (affects strategy selection)
+/// - `tip_height`: The current blockchain height (used as the base for time locks)
+/// - `acc_locktime`: The accumulated for the absolute locktime.
```
evanlinjin:
```suggestion
/// - `acc_locktime`: The accumulated absolute locktime.
```

src/utils.rs
```diff
@@ -87,45 +114,53 @@ pub fn apply_anti_fee_sniping(
...
-        tx.lock_time = new_locktime;
+        tx.lock_time = if acc_locktime.is_implied_by(afs_locktime) {
+            afs_locktime
+        } else {
+            acc_locktime
+        };
     } else {
         // Use Sequence
-        tx.lock_time = LockTime::ZERO;
+        tx.lock_time = acc_locktime;
```
evanlinjin: Nit: This is already set in `create_psbt`. Redundant.

src/utils.rs
```diff
@@ -64,20 +89,22 @@ use rand_core::RngCore;
 pub fn apply_anti_fee_sniping(
     tx: &mut Transaction,
     inputs: &[Input],
-    current_height: absolute::Height,
-    rbf_enabled: bool,
+    tip_height: absolute::Height,
+    acc_locktime: absolute::LockTime,
```
evanlinjin: I just realized `tx.lock_time` is the accumulated locktime. We can remove `acc_locktime` now.

src/utils.rs
```diff
@@ -1,20 +1,48 @@
...
+/// Error returned by [`apply_anti_fee_sniping`].
+#[derive(Debug, Clone)]
+pub enum AntiFeeSnipingError {
+    /// Transaction `version` must be >= 2 for AFS to use relative locktimes.
+    UnsupportedVersion(Version),
+    /// AFS only supports height-based locktimes. A time-based (MTP) locktime
+    /// — either from `min_locktime` or from an input's CLTV requirement —
+    /// cannot be combined with the height semantics used here.
+    UnsupportedLockTime(absolute::LockTime),
+}
```
evanlinjin: we need to actually expose this.

In `lib.rs`:

```rust
pub use utils::*;
```

evanlinjin (changes requested):
Make `apply_anti_fee_sniping` explicitly private!

examples/anti_fee_sniping.rs
```diff
@@ -88,13 +88,12 @@ fn main() -> anyhow::Result<()> {
                 },
             )?;
 
-        let fallback_locktime: LockTime = LockTime::from_consensus(tip_height.to_consensus_u32());
+        let min_locktime: LockTime = LockTime::from_consensus(tip_height.to_consensus_u32());
 
         let selection_inputs = selection.inputs.clone();
 
         let psbt = selection.create_psbt(PsbtParams {
-            enable_anti_fee_sniping: true,
```
evanlinjin, line 96: Don't we need to enable AFS for this example to make sense?

evanlinjin:
We are lacking commit messages, PR description has factual errors, missing `!` to signal breaking change.

I'll push it forward!

evanlinjin:
@aagbotemi I pushed a Claude-generated test commit on top, let me know what you think.

src/selection.rs
```diff
@@ -632,6 +573,150 @@ mod tests {
...
+                .find(|t| t.previous_output == csv_outpoint)
+                .expect("csv input must be present");
+            assert_eq!(
+                csv_txin.sequence, csv_sequence,
+                "AFS must not overwrite the sequence of a CSV-bearing Taproot input",
+            );
+            assert!(tx
+                .input
+                .iter()
+                .any(|t| t.previous_output == regular_outpoint));
```
aagbotemi: Is there any reason for adding this? I suggest removing it for clarity.

evanlinjin: Good point, the AI was hallucinating. Instead, we should check whether the non-CSV input was picked and AFS is enforced via it's sequence value (if it went down the sequence-based AFS path).

evanlinjin: Addressed. Looked through the other tests as well more thoroughly.

evanlinjin (approved):
ACK 5df9bb1c054cdcc825a0b0f5894d63a2cd8b6c91

src/utils.rs
```diff
@@ -87,45 +105,44 @@ pub fn apply_anti_fee_sniping(
             let input = inputs
                 .iter()
                 .find(|input| input.prev_outpoint() == txin.previous_output)?;
-            if input.prev_txout().script_pubkey.is_p2tr() {
+            if input.prev_txout().script_pubkey.is_p2tr() && input.relative_timelock().is_none() {
```
ValuedMammal, line 108: BIP326 is under-specified when it comes to inputs that already come with a CSV attached. We chose a conservative reading by excluding all inputs with a relative timelock, but a refinement could be to admit block-based CSV inputs whose requirement is dominated by their confirmation depth.

Then eligibility becomes "is p2tr, and either has no relative timelock, or has a block-based one whose value is `< input.confirmations(tip_height)`" at which point the logic collapses to just `confirmations` (except that the 10% random backoff must floor at `max(csv_blocks, 1)` instead of 1).

src/utils.rs
```diff
@@ -87,45 +105,44 @@ pub fn apply_anti_fee_sniping(
...
-        let new_locktime = LockTime::from_height(locktime).expect("must be valid Height");
+        let afs_locktime = LockTime::from_height(afs_height).expect("must be valid Height");
 
-        tx.lock_time = new_locktime;
+        if tx.lock_time.is_implied_by(afs_locktime) {
+            tx.lock_time = afs_locktime;
+        }
     } else {
         // Use Sequence
-        tx.lock_time = LockTime::ZERO;
```
ValuedMammal, line 125: The sequence branch exists so your transaction resembles an off-chain settlement spending a timelock path, and those transactions have nLockTime = 0 with a CSV-driven sequence. A transaction carrying both a near-tip locktime and a confirmation-depth sequence resembles neither an ordinary AFS wallet spend nor a contract closing. You've built a third fingerprint, which is worse than either branch alone. So the current code isn't just semantically loose, it actively undercuts the branch's only purpose. The sequence branch is only ever coherent when tx.lock_time is already zero, so make that a precondition as part of `must_use_locktime`.

```rust
let must_use_locktime = tx.lock_time != LockTime::ZERO
    || taproot_inputs.is_empty()
    || inputs.iter().any(|input| { /* ... */ });
```

src/selection.rs
```diff
@@ -27,15 +24,17 @@ pub struct PsbtParams {
...
     ///
-    /// The locktime to use if no input specifies a required absolute locktime.
-    ///
-    /// It is best practice to set this to the latest block height to avoid fee sniping.
-    pub fallback_locktime: absolute::LockTime,
+    /// The final `tx.lock_time` is the maximum of this value and any absolute locktime required by
+    /// an input's CLTV, provided the locktime units agree. If `min_locktime` uses a different unit
+    /// (block-height vs. time) than an input's CLTV, it is ignored — a height-based `min_locktime`
+    /// will not be combined with a time-based CLTV (and vice versa).
+    pub min_locktime: absolute::LockTime,
```
ValuedMammal, line 33: There is essentially no value of `min_locktime` that is simultaneously meaningful and compatible with AFS. That's a strong signal the two fields are answering the same question - "who decides nLockTime when the inputs don't force it, and should both be settable?"

- Below `tip - 100`: AFS always wins the max, so the floor does nothing
- Within `(tip - 100, tip]`: The conflict zone, truncates or erases the 10% backoff
- Above tip: The floor always wins, so AFS does nothing

So I'd fold them into a `LockTimePolicy` with the default as `Min(LockTime::ZERO)`, replacing both PsbtParams fields. Input CLTV requirements still max on top in either arm; document that once in the enum's doc, instead of scattered across two fields that have to cross-reference each other.

```rust
pub enum LockTimePolicy {
    /// Floor `tx.lock_time` at this value.
    Min(absolute::LockTime),
    /// Apply BIP-326 anti-fee-sniping using this chain tip.
    AntiFeeSniping(absolute::Height),
}
```

PR #66, opened 2026-05-13
feat(input): add set_sequence for per-input sequence override

### LINKED ISSUE #61: add `Input::set_sequence` for per-input sequence override
### Description

`Input` currently has no way for the caller to set a sequence value after construction. The sequence is either:

* **`Plan` variant:** Derived from `plan.relative_timelock` (which can be `None`).
* **`PsbtInput` variant:** Taken from `psbt::Input.sequence`.

Neither path lets the caller say "use this specific sequence value" after-the-fact.

### Motivation

Several workflows want per-input sequence control:

1. BIP326 AFS (sequence branch) needs to override one randomly chosen taproot input's sequence.
2. Atomic swaps / coordinated multiparty txs sometimes need a specific sequence on a specific input.

These values cannot be provided by a `Plan`, and `PsbtParams::fallback_sequence` applies to all inputs and cannot override a sequence value already suggested by a `Plan`.

### Proposed API

```rust
impl Input {
    /// Override the sequence value this input contributes to the
    /// resulting tx.
    ///
    /// # Errors
    ///
    /// Returns `SetSequenceError::IncompatibleRelativeTimelock` if the
    /// new value wouldn't satisfy the input's existing relative-timelock
    /// requirement (BIP68/CSV).
    pub fn set_sequence(&mut self, sequence: Sequence) -> Result<(), SetSequenceError>;
}

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum SetSequenceError {
    IncompatibleRelativeTimelock {
        required: relative::LockTime,
        new: Sequence,
    },
}
```

A new field will need to be added to `PlanOrPsbtInput::Plan`.

```rust
enum PlanOrPsbtInput {
    Plan {
        plan: Box<Plan>,
        /// New field.
        sequence_override: Option<Sequence>,
    }
}
```

`Input::sequence()` will need to be modified to return the `sequence_override` value if it exists.

aagbotemi:
### Description

Add `Input::set_sequence` for per-input `nSequence` manipulation as an alternative to `PsbtParams::fallback_sequence` (global `nSequence` fallback setting). Also allow psbt-inputs to convey CLTV intent.

Closes #61 

### Changelog Notice

```md
Added:
- `Input::set_sequence` and `SetSequenceError` for per-input sequence override.

Removed:
- Removed `PsbtParams::fallback_sequence` field in favor of per-input `nSequence` manipulation.

Changed:
- `Input::from_psbt_input()` has a new `absolute_timelock` parameter which communicates the intended spend path's CLTV value.
- Default `nSequence` value for inputs is now ENABLE_RBF_NO_LOCKTIME (was ENABLE_LOCKTIME_NO_RBF).
```

### Checklist

- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
- [x] I've added tests for the new feature
- [x] I've added docs for the new feature
- [x]  I'm linking the issue being fixed by this PR

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -37,13 +37,61 @@ impl ConfirmationStatus {
     }
 }
 
+/// Error returned by [`Input::set_sequence`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SetSequenceError {
+    /// The new sequence value does not satisfy the input's relative-timelock requirement.
+    RelativeTimelockNotSatisfied {
+        /// The relative timelock required by the input's plan.
+        required: relative::LockTime,
+        /// The sequence value the caller attempted to set.
+        new: Sequence,
+    },
+    /// The input executes CLTV (absolute timelock), but `Sequence::MAX` disables `nLockTime`,
+    /// which would cause CLTV to always fail at script execution.
+    AbsoluteTimelockDisabled {
+        /// The absolute timelock that will be executed by this input.
+        required: absolute::LockTime,
+    },
+}
+
+impl fmt::Display for SetSequenceError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::RelativeTimelockNotSatisfied { required, new } => write!(
+                f,
+                "sequence {} does not satisfy required relative timelock {}",
+                new, required,
+            ),
+            Self::AbsoluteTimelockDisabled { required } => write!(
+                f,
+                "Sequence::MAX disables nLockTime, but this input executes CLTV with absolute timelock {}",
+                required,
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for SetSequenceError {}
+
+/// Internal representation of an input's spending data.
 #[derive(Debug, Clone)]
 enum PlanOrPsbtInput {
-    Plan(Box<Plan>),
+    /// Input described by a miniscript [`Plan`].
+    Plan {
+        /// The plan describing how to satisfy this input.
+        plan: Box<Plan>,
+        /// Sequence override.
+        ///
+        /// Takes precedence over the sequence implied by `plan.relative_timelock`
+        /// when computing [`Input::sequence`].
+        sequence_override: Option<Sequence>,
+    },
     PsbtInput {
         psbt_input: Box<psbt::Input>,
         sequence: Sequence,
-        absolute_timelock: absolute::LockTime,
+        absolute_timelock: Option<absolute::LockTime>,
         satisfaction_weight: usize,
     },
 }
@@ -56,6 +104,7 @@ impl PlanOrPsbtInput {
         sequence: Sequence,
         psbt_input: psbt::Input,
         satisfaction_weight: usize,
+        absolute_timelock: Option<absolute::LockTime>,
     ) -> Result<Self, FromPsbtInputError> {
         // We require at least one of the witness or non-witness utxo
         if psbt_input.witness_utxo.is_none() && psbt_input.non_witness_utxo.is_none() {
@@ -64,14 +113,14 @@ impl PlanOrPsbtInput {
         Ok(Self::PsbtInput {
             psbt_input: Box::new(psbt_input),
             sequence,
-            absolute_timelock: absolute::LockTime::ZERO,
+            absolute_timelock,
             satisfaction_weight,
         })
     }
 
     pub fn plan(&self) -> Option<&Plan> {
         match self {
-            PlanOrPsbtInput::Plan(plan) => Some(plan),
+            PlanOrPsbtInput::Plan { plan, .. } => Some(plan),
             _ => None,
         }
     }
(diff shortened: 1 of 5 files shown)


evanlinjin (changes requested):
Looks good to me.

Can you create another commit to remove `PsbtParams::fallback_sequence`? For inputs which don't have a sequence value provided by plan or otherwise, let's use `Sequence::ENABLE_RBF_NO_LOCKTIME` as per Bitcoin Core.

Rationale: A custom global fallback sequence value is useless. Callers will always want per-input control.

aagbotemi:
Thank you for the review

evanlinjin (approved):
ACK 5f2823243e8149e37a931532f8a17b4ed354788c

src/input.rs
```diff
@@ -510,6 +564,44 @@ impl Input {
...
+    ///
+    /// Returns [`SetSequenceError::IncompatibleRelativeTimelock`] if the input has a
+    /// plan-required relative timelock and the new sequence does not satisfy it
+    /// (BIP-68 / OP_CSV).
+    pub fn set_sequence(&mut self, sequence: Sequence) -> Result<(), SetSequenceError> {
+        match &mut self.plan {
+            PlanOrPsbtInput::Plan {
+                plan,
+                sequence_override,
+            } => {
```
evanlinjin, line 632: There is one other check that I forgot to mention.

In [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/82733e61deea24cfd344aed36020675a15cc672e/src/script/interpreter.cpp#L1745-L1779), if the input script executes CLTV, then `input.sequence` cannot be `Sequence::MAX`.

Probably need a new error variant called `IncompatibleAbsoluteTimelock` or something.

evanlinjin (changes requested):

src/input.rs
aagbotemi: Done in f80556db1fc1dfb6311680df8bc6f19638db65f8

evanlinjin: What you've done in f80556db1fc1dfb6311680df8bc6f19638db65f8 has value, but it is NOT what I requested and lacks documentation.

### My initial suggestion

Introduces a check on `set_sequence` and a subsequent error variant on `SetSequenceError`.

### To move forward

1. Add checks to `set_sequence`.
  * `SetSequenceError` should have a new variant `IncompatibleAbsoluteTimelock`.
  * **For plan-based branch** - We know whether CLTV will be executed from `plan.absolute_timelock`. If so, setting `Sequence::MAX` should error.
  * **For psbt-based branch** - If `absolute_timelock` is stored as `Some` and we are setting `Sequence::MAX`, this should error.

2. Fix f80556db1fc1dfb6311680df8bc6f19638db65f8
  * The commit allowed psbt-inputs to contribute to `acc_locktime`. This should be documented in the commit message and PR description.
  * Improve the docs. Currently, it's unclear what the new `Option<absolute::LockTime>` parameter does. The following should be mentioned:
    * Pass `None` if the input's spending path does not execute CLTV.
    * If `Some`, this value contributes to `tx.nLockTime` and is validated against `Sequence::MAX`.
    * If you pass `None` for a script that does execute CLTV, the resulting tx will be consensus-invalid — this is the caller's responsibility."

evanlinjin (changes requested):

src/input.rs
```diff
@@ -37,13 +37,63 @@ impl ConfirmationStatus {
...
+    },
+    /// The input executes CLTV (absolute timelock), but `Sequence::MAX` disables `nLockTime`,
+    /// which would cause CLTV to always fail at script execution.
+    IncompatibleAbsoluteTimelock {
+        /// The absolute timelock that will be executed by this input.
+        required: absolute::LockTime,
+        /// The sequence value the caller attempted to set.
+        new: Sequence,
+    },
+}
```
evanlinjin, line 56: A gripe with the variant naming (sorry, `IncompatibleRelativeTimelock` was my suggestion).

- **`IncompatibleRelativeTimelock` -> `RelativeTimelockNotSatisfied`**
  Technically, it's the `nSequence` value that is incompatible, but we can't have `IncompatibleSequence` since both variants are technically "incompatible sequence".

- **`IncompatibleAbsoluteTimelock` -> `AbsoluteTimelockDisabled` (but with the `new: Sequence` field removed)**
  BIP-65's CLTV failure only happens with `Sequence::MAX` which signals for disabling OP_CLTV.

evanlinjin (changes requested):

src/input.rs
```diff
@@ -166,6 +223,14 @@ pub enum FromPsbtInputError {
     InvalidOutPoint(OutPoint),
     /// The input's UTXO is missing or invalid
     UtxoCheck,
+    /// Input uses CLTV (absolute timelock) but sequence is `Sequence::MAX`,
+    /// which disables locktime and causes CLTV to fail.
+    IncompatibleAbsoluteTimelock {
```
evanlinjin: For consistency:

```suggestion
    AbsoluteTimelockDisabled {
```

src/input.rs
aagbotemi: I've fixed it

evanlinjin:
@aagbotemi Thanks for working on this PR. The code looks good. However, some commits forgot to convey that they are breaking changes, changes are mixed between commits, the changelog notice is incomplete, docs have broken links.

I'll take over this PR for final touches.

evanlinjin (approved):
ACK 87bccd85c42bf30a18a3467d5972a4d1191fe447

PR #69, opened 2026-05-14
Introduce per-input sighash type via `Input::set_sighash_type`

### LINKED ISSUE #60: `PsbtParams::sighash_type` does not make sense
### Description

The tx-level `PsbtParams::sighash_type` field is dead weight: its default of `None` covers the `SIGHASH_ALL` case (the 99% common one), and the rare callers who'd want to override usually want **per-input** control (atomic swaps, coordinated txs, ANYONECANPAY workflows). It also has an awkward asymmetry: it only applies to Plan-variant inputs, while `PsbtInput`-variant inputs already carry their own sighash in `psbt::Input::sighash_type`.

Just removing the field is most of the fix. The trickier question is what `create_psbt` should do for Plan-variant inputs by default, because Taproot signature size depends on sighash (64 bytes for `SIGHASH_DEFAULT`, 65 for anything else). Miniscript's `Plan` bakes this assumption in via `TaprootCanSign::sighash_default`, and `Plan::satisfaction_weight()` is only accurate if the signer actually uses the assumed sighash. If `create_psbt` leaves sighash entirely up to the caller, the catastrophic case is: Plan was built with `sighash_default = true` (64-byte sigs assumed), signer produces a 65-byte signature, transaction is silently under-funded.

### Proposed Changes

1. **Remove `PsbtParams::sighash_type`.**
2. **In `create_psbt`, auto-write `TapSighashType::Default` only when the Plan's Schnorr placeholders are *uniformly* 64 bytes (all-Default-Taproot).** This is the one case where the Plan's weight assumption forces a specific sighash, and where a misbehaving (or BIP-defaulted) signer producing 65-byte sigs would silently under-fund the tx. Every other case is left unset:

   | Plan's Taproot placeholder sizes | Behavior |
   |---|---|
   | All 64B (uniform Default-Taproot) | Auto-write `TapSighashType::Default`. |
   | All 65B (uniform non-Default-Taproot) | Leave unset. Signer defaults to `SIGHASH_ALL` (65B — matches budget). |
   | Mixed 64B + 65B | Leave unset. Plan weight is accurate only if signers coordinate externally to use the per-key declared sighashes; PSBT cannot represent per-key sighash intent. Documented caveat, not an error — the Plan is consensus- and miniscript-valid. |
   | No Schnorr placeholders (ECDSA) | Leave unset. Sighash-invariant for weight. |

   Sizes are read from `Plan::witness_template()` (the trailing `usize` on each `SchnorrSigPk`/`SchnorrSigPkHash` placeholder).

3. **Callers needing non-`ALL` semantics** (`SINGLE`, `NONE`, `*_ANYONECANPAY`) set `psbt.inputs[i].sighash_type` directly after `create_psbt` returns. `psbt::Input.sighash_type` becomes the single source of truth for per-input sighash on the constructed PSBT — uniform across Plan-derived and PSBT-derived inputs.

The auto-write of `Default` for uniform-64B Plans guards the catastrophic direction (64B-budget, 65B-signature) by default. Overriding it post-construction is explicit caller intent — to be documented on `create_psbt` as the one remaining footgun (mismatched byte length silently mis-funds the tx).

Design discussion (originally on the closed #69): bitcoindevkit/bdk-tx#69 (comment).

evanlinjin:
### Description

Closes #60

Replace `PsbtParams::sighash_type` with a per-input setter/getter on `Input`. `Selection::create_psbt` propagates each input's sighash type into the resulting PSBT input.

### Changelog notice

```md
Added:
- `Input::set_sighash_type` (setter) and `Input::sighash_type` (getter) methods which specifies per-input sighash type.

Removed:
- `PsbtParams::sighash_type` which sets sighash type for all inputs. Replaced by `Input::{set_}sighash_type` for per-input control.
```

### Before submitting

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -87,6 +87,7 @@ enum PlanOrPsbtInput {
         /// Takes precedence over the sequence implied by `plan.relative_timelock`
         /// when computing [`Input::sequence`].
         sequence_override: Option<Sequence>,
+        sighash_type: Option<psbt::PsbtSighashType>,
     },
     PsbtInput {
         psbt_input: Box<psbt::Input>,
@@ -153,11 +154,19 @@ impl PlanOrPsbtInput {
             PlanOrPsbtInput::Plan {
                 plan,
                 sequence_override,
+                ..
             } => sequence_override.or_else(|| {
                 plan.relative_timelock
                     .map(|relative_timelock| relative_timelock.to_sequence())
             }),
             PlanOrPsbtInput::PsbtInput { sequence, .. } => Some(*sequence),
+        }
+    }
+
+    pub fn sighash_type(&self) -> Option<psbt::PsbtSighashType> {
+        match self {
+            PlanOrPsbtInput::Plan { sighash_type, .. } => *sighash_type,
+            PlanOrPsbtInput::PsbtInput { psbt_input, .. } => psbt_input.sighash_type,
         }
     }
 
@@ -291,6 +300,7 @@ impl Input {
             plan: PlanOrPsbtInput::Plan {
                 plan: Box::new(plan),
                 sequence_override: None,
+                sighash_type: None,
             },
             status,
             is_coinbase,
@@ -312,6 +322,7 @@ impl Input {
             plan: PlanOrPsbtInput::Plan {
                 plan: Box::new(plan),
                 sequence_override: None,
+                sighash_type: None,
             },
             status,
             is_coinbase,
@@ -598,6 +609,26 @@ impl Input {
         self.plan.sequence()
     }
 
+    /// Sighash type.
+    pub fn sighash_type(&self) -> Option<psbt::PsbtSighashType> {
+        self.plan.sighash_type()
+    }
+
+    /// Set the sighash type for this input.
+    ///
+    /// This overrides any sighash type already attached to the input — whether carried in via
+    /// [`Input::from_psbt_input`] (as [`psbt::Input::sighash_type`]) or set by a prior call.
+    ///
+    /// Accepts anything convertible into [`psbt::PsbtSighashType`], including the standard
+    /// `EcdsaSighashType` and `TapSighashType` from `rust-bitcoin`.
+    pub fn set_sighash_type(&mut self, sighash_type: impl Into<psbt::PsbtSighashType>) {
+        let new = Some(sighash_type.into());
+        match &mut self.plan {
+            PlanOrPsbtInput::Plan { sighash_type, .. } => *sighash_type = new,
+            PlanOrPsbtInput::PsbtInput { psbt_input, .. } => psbt_input.sighash_type = new,
+        }
+    }
+
     /// The weight in witness units needed for satisfying the [`Input`].
     ///
     /// The satisfaction weight is the combined size of the fully satisfied input's witness
@@ -629,6 +660,7 @@ impl Input {
             PlanOrPsbtInput::Plan {
                 plan,
                 sequence_override,
+                ..
             } => {
                 if let Some(required) = plan.absolute_timelock {
                     if sequence == Sequence::MAX {
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -39,14 +39,6 @@ pub struct PsbtParams {
     /// [`non_witness_utxo`]: bitcoin::psbt::Input::non_witness_utxo
     pub mandate_full_tx_for_segwit_v0: bool,
 
-    /// Sighash type to be used for each input.
-    ///
-    /// This option only applies to [`Input`]s that include a plan, as otherwise the given PSBT
-    /// input can be expected to set a specific sighash type. Defaults to `None` which will not
-    /// set an explicit sighash type for any input. (In that case the sighash will typically
-    /// cover all of the outputs).
-    pub sighash_type: Option<bitcoin::psbt::PsbtSighashType>,
-
     /// Apply BIP-326 anti-fee-sniping (AFS) protection, using the given block height.
     ///
     /// * `None` (default) — no AFS is applied.
@@ -84,7 +76,6 @@ impl Default for PsbtParams {
             version: transaction::Version::TWO,
             min_locktime: absolute::LockTime::ZERO,
             mandate_full_tx_for_segwit_v0: true,
-            sighash_type: None,
             anti_fee_sniping: None,
         }
     }
@@ -253,7 +244,7 @@ impl Selection {
                     }
                 }
 
-                psbt_input.sighash_type = params.sighash_type;
+                psbt_input.sighash_type = plan_input.sighash_type();
 
                 continue;
             }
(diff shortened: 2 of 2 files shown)


src/input.rs
```diff
@@ -598,6 +609,28 @@ impl Input {
...
+    /// `EcdsaSighashType` and `TapSighashType` from `rust-bitcoin`.
+    pub fn set_sighash_type(&mut self, sighash_type: impl Into<psbt::PsbtSighashType>) {
+        let sighash_type = Some(sighash_type.into());
+        match &mut self.plan {
+            PlanOrPsbtInput::Plan {
+                sighash_type: sh, ..
+            } => *sh = sighash_type,
+            PlanOrPsbtInput::PsbtInput { psbt_input, .. } => psbt_input.sighash_type = sighash_type,
+        }
+    }
```
noahjoeris, line 630: Nit: This is a bit hard to read.
```suggestion
    pub fn set_sighash_type(&mut self, sighash_type: impl Into<psbt::PsbtSighashType>) {
        let new = Some(sighash_type.into());
        match &mut self.plan {
            PlanOrPsbtInput::Plan { sighash_type, .. } => *sighash_type = new,
            PlanOrPsbtInput::PsbtInput { psbt_input, .. } => psbt_input.sighash_type = new,
        }
    }
```

noahjoeris (approved):
Nice one!
ACK 9ffb42c

src/input.rs
evanlinjin: Changed.

noahjoeris (approved):
ACK 95d4b35

evanlinjin:
Two problems with this PR right now.

1. A transaction input can have multiple signatures (and therefore multiple sighashes with different sighash types).
2. The `Input::sighash_type` can actually disagree with the `Plan`. This matters because for Taproot signatures, DEFAULT and ALL sighash flags will produce signatures of different sizes, affecting coin selection.

## Why does a PSBT Input have a single Sighash Type field?

This seems counter-intuitive since a finalized input can have multiple signatures and those signatures are not required to have uniform sighash types.

BIP-174 describes the reasoning clearly:

> The 32-bit unsigned integer specifying the sighash type to be used for this input. Signatures for this input must use the sighash type, finalizers must fail to finalize inputs which have signatures that do not match the specified sighash type. Signers who cannot produce signatures with the sighash type must not provide a signature. 

In fact, this field is not required at all!

> If a sighash type is not provided, the signer should sign using SIGHASH_ALL, but may use any sighash type they wish.

## Proposed Solution

*  Revert commit 95d4b35eb48f45779180a6c6e938878f027023e5 (drop `set_sighash_type` / `sighash_type` entirely).
* In `Selection::create_psbt`:
  * For psbt-`Input` - set PSBT Sighash Type field to whatever defined in the input.
  * For plan`Input` - only set PSBT Sighash Type if all signature plans require the DEFAULT sighash flag.

We need to make changes upstream (to the `miniscript` repo):

```rust
impl Plan {
    /// Returns `Some(true)` if every Schnorr placeholder assumes SIGHASH_DEFAULT,
    /// `Some(false)` if every one assumes non-Default,
    /// `None` for non-Taproot plans or mixed plans (which can't be expressed by a single sighash policy).
    pub fn tap_sighash_default(&self) -> Option<bool>
}
```

## AI Curation of Proposed Solution 

image

This needs more thought....

evanlinjin:
**Updated design for this PR.**

After working through it, the per-Input sighash API turns out to be the source of the problem rather than a feature. Removing it and letting `psbt::Input.sighash_type` be the single source of truth for per-input sighash policy.

## The problem

For Taproot, signature size depends on sighash: 64 bytes for `SIGHASH_DEFAULT`, 65 bytes for anything else. Miniscript's `Plan` bakes this assumption in at construction time (via `TaprootCanSign::sighash_default`), and `Plan::satisfaction_weight()` is only accurate if the signer actually uses the assumed sighash.

`Input::set_sighash_type` introduces sighash state on `Input` independent of the Plan it was built from, so the two can disagree. The disagreement is silent — weight estimates become wrong, and coin selection drifts off.

## Why sighash belongs at the PSBT layer, not on `Input`

BIP-174 treats sighash as a per-PSBT-input policy:

> *"The 32-bit unsigned integer specifying the sighash type to be used for this input. Signatures for this input must use the sighash type, finalizers must fail to finalize inputs which have signatures that do not match the specified sighash type."*

And when unset:

> *"If a sighash type is not provided, the signer should sign using SIGHASH_ALL, but may use any sighash type they wish."*

One input, one sighash policy, enforced at finalization. `Input` in bdk-tx describes *what is being spent and how it can be satisfied*. The sighash policy is a property of the constructed PSBT, not of the spend description — so it belongs on the PSBT input itself (`psbt::Input.sighash_type`), which rust-bitcoin already exposes as a public, mutable field.

## Proposed design

**1. Revert the per-Input sighash API.** Drop `Input::set_sighash_type` and `Input::sighash_type`. `Input` carries no sighash state.

**2. In `create_psbt`, auto-write the PSBT input's `sighash_type` only where the Plan's weight assumption forces a specific value — for Default-Taproot Plans.** Everything else is left unset; the caller mutates `psbt.inputs[i].sighash_type` directly if they need a specific non-`ALL` sighash.

The witness-size assumption is read from `Plan::witness_template()` via a local helper `tap_sighash_default(plan) -> Option<bool>`: the trailing `usize` on the first `SchnorrSigPk`/`SchnorrSigPkHash` placeholder is `64` (default-Taproot) or `65` (non-default), returning `None` for ECDSA Plans. Mixed sizes within a single Plan are pathological — debug-assert uniformity.

| `tap_sighash_default(plan)` | Behavior in `create_psbt` |
|---|---|
| `Some(true)` (Default-Taproot Plan) | Auto-write `TapSighashType::Default`. This is the one case where the Plan's 64-byte weight assumption locks the sighash to exactly one value; any other choice silently under-funds the transaction. |
| `Some(false)` (non-Default-Taproot Plan) | Leave the field unset. Signer defaults to `SIGHASH_ALL` (65 bytes — matches the Plan's weight assumption). Callers needing non-`ALL` semantics (`SINGLE`, `NONE`, `*_ANYONECANPAY`) set `psbt.inputs[i].sighash_type` post-construction. |
| `None` (ECDSA — sighash-invariant for weight) | Leave the field unset. |

PSBT-based `Input`s (built via `Input::from_psbt_input`) are unaffected — their embedded `psbt::Input` already owns its `sighash_type` field. The auto-write applies only to Plan-based inputs.

**3. Doc the one remaining footgun on `create_psbt`:** mutating `psbt.inputs[i].sighash_type` after construction on a Taproot input is allowed but the caller is responsible for keeping the byte length consistent with the Plan's witness-size assumption (64B for `Default`, 65B for any other Taproot sighash). Mismatch silently mis-funds the transaction. The auto-write of `Default` for Default-Taproot Plans guards the catastrophic direction (64B-budget, 65B-signature) by default; overriding it is explicit caller intent.

## Net effect

- `Input` is purely a spend description, cannot carry inconsistent sighash state.
- Single source of truth for per-input sighash on the constructed PSBT: `psbt.inputs[i].sighash_type`. No parallel override API.
- Default-Taproot Plans get safe auto-lock with zero caller ceremony — prevents the catastrophic 64B→65B underpayment.
- Non-Default-Taproot Plans match BIP-174's implicit default (`SIGHASH_ALL`, 65 bytes) and stay weight-accurate; callers needing non-`ALL` semantics make a one-line post-construction edit.
- ECDSA Plans are unconstrained, matching reality.

nymius (commented):
Is this PR complete? I agree with the code here, but I wouldn't relay on a witness guessing scheme for extracting the SIGHASH type in the `Plan` variant.

evanlinjin:
> Is this PR complete? I agree with the code here, but I wouldn't relay on a witness guessing scheme for extracting the SIGHASH type in the `Plan` variant.

It is not complete, waiting on upstream changes. Where did you think we were guessing the sighash type?

Sometimes plan is based on Sighash DEFAULT - if the resultant tx uses anything else, the feerate/fee will be different to what CS assumed. The idea is to error on inconsistency.

nymius:
I was thinking in the `rust-miniscript` change. In the light of your last response in that PR, I guess miniscript doesn't expose the sighash values because is presuming the descriptor holder knows which sighashes it's going to use. Can't we avoid bubbling up the CanSign bool by getting the value earlier?

evanlinjin:
> I was thinking in the `rust-miniscript` change. In the light of your last response in that PR, I guess miniscript doesn't expose the sighash values because is presuming the descriptor holder knows which sighashes it's going to use. Can't we avoid bubbling up the CanSign bool by getting the value earlier?

I realized that we can actually use `Plan::witness_template` method to check sighash constraints directly. So no need for upstream changes to `miniscript`.

evanlinjin:
Closing. Rationale: https://github.com/bitcoindevkit/bdk-tx/pull/69#issuecomment-4469793461

PR #72, opened 2026-05-17
refactor(selection)!: encapsulate Selection and fail-fast on mixed-unit locktimes

### LINKED ISSUE #58: `Selection` is constructable with mixed-unit absolute locktimes
### The Problem

Currently, this error is surfaced during `Selection::create_psbt` (as `CreatePsbtError::LockTypeMismatch`). However, this error should be detected earlier in the pipeline - `Selection` is technically invalid with locktime type mismatch.

### Proposed Solution

* `Selector` shouldn't be constructable with `InputCandidates` with inputs that have different locktime-type requirements. `Selector::new` should error (with a new `SelectorError::LockTypeMismatch` variant).

* External crates should not be able to bypass the check done in `Selector::new`. We should make `Selection` only constructable via a `Selector`. Additionally, make all fields of `Selection` private and add getter methods so they cannot be modified directly by the caller.

* `CreatePsbtError::LockTypeMismatch` can be removed after the above changes.

### LINKED ISSUE #67: Rename `utils.rs` to `afs.rs`?
Can we rename this module to `afs.rs`?

_Originally posted by @evanlinjin in https://github.com/bitcoindevkit/bdk-tx/pull/65#discussion_r3234063110_

evanlinjin:
### Description

Detects mixed-unit absolute timelock requirements at `Selector::new` instead of at PSBT creation, and prevents external code from constructing or rearranging a `Selection` in ways that would bypass the check.

Closes #58

### Changes

1. **Fail-fast in the selector.** `Selector::new` now returns `SelectorError::LockTypeMismatch` when candidate inputs mix height-based and time-based absolute timelocks. The corresponding `CreatePsbtError::LockTypeMismatch` is removed and `Selection::create_psbt`'s internal locktime accumulator becomes infallible (with a `debug_assert!` guarding the upstream invariant).

2. **Encapsulate `Selection`.** `Selection.inputs` / `Selection.outputs` are now private. Access is via `inputs()` / `outputs()`. Per-input mutation goes through a new `InputMut<'_>` handle (returned by `input_mut(outpoint)` / `inputs_mut()`) which derefs to `&Input` for reads and exposes only `set_sequence`, preventing whole-input replacement that would silently break coin-selection invariants.

3. **Reorder methods.** Adds `sort_inputs_by`, `shuffle_inputs`, `sort_outputs_by`, `shuffle_outputs` on `Selection` so callers can apply BIP-69 or chain-analysis-resistant ordering without raw mutable slice access.

### Scope creep

Bundled in this PR (let me know if you'd prefer these split out):

- **Workspace refactor.** `src/utils.rs` is split into `src/afs.rs` (AFS types) and `src/no_std_rand.rs` (no-std rand helpers). `no_std_rand` is now crate-private since it has no external consumers.
- **`fisher_yates_shuffle` helper.** Added to `no_std_rand` and used to back `Selection::shuffle_inputs` / `shuffle_outputs`.

### Changelog Notice

```md
Added:
- `SelectorError::LockTypeMismatch` — raised by `Selector::new` when candidate inputs mix height-based and time-based absolute timelocks.
- `Selection::inputs` / `Selection::outputs` accessors (replacing the public fields).
- `Selection::input_mut` / `Selection::inputs_mut` returning a new `InputMut<'_>` handle that only permits `set_sequence`.
- `Selection::sort_inputs_by`, `Selection::shuffle_inputs`, `Selection::sort_outputs_by`, `Selection::shuffle_outputs` for reordering without exposing raw mutable access.

Removed:
- `CreatePsbtError::LockTypeMismatch` — superseded by `SelectorError::LockTypeMismatch`.

Changed:
- `Selection.inputs` and `Selection.outputs` are no longer public fields. External construction of `Selection` is no longer possible; obtain one from `Selector::try_finalize`.
```

### Checklist

- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
- [x] I've added tests for the new feature
- [x] I've added docs for the new feature
- [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -665,6 +665,39 @@ impl Input {
     }
 }
 
+/// Mutable handle to an [`Input`] held inside a [`Selection`].
+///
+/// Returned by [`Selection::input_mut`] and [`Selection::inputs_mut`]. This wrapper restricts
+/// mutation to operations that preserve [`Selection`]'s coin-selection invariants.
+///
+/// Read-only access to the underlying [`Input`] is available via [`Deref`].
+///
+/// [`Selection`]: crate::Selection
+/// [`Selection::input_mut`]: crate::Selection::input_mut
+/// [`Selection::inputs_mut`]: crate::Selection::inputs_mut
+/// [`Deref`]: core::ops::Deref
+#[derive(Debug)]
+pub struct InputMut<'a>(&'a mut Input);
+
+impl<'a> InputMut<'a> {
+    pub(crate) fn new(input: &'a mut Input) -> Self {
+        Self(input)
+    }
+
+    /// See [`Input::set_sequence`].
+    pub fn set_sequence(&mut self, sequence: Sequence) -> Result<(), SetSequenceError> {
+        self.0.set_sequence(sequence)
+    }
+}
+
+impl core::ops::Deref for InputMut<'_> {
+    type Target = Input;
+
+    fn deref(&self) -> &Input {
+        self.0
+    }
+}
+
 /// Input group. Cannot be empty.
 #[derive(Debug, Clone)]
 pub struct InputGroup(Vec<Input>);
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -1,5 +1,6 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::cmp::Ordering;
 use core::fmt::{Debug, Display};
 
 use miniscript::bitcoin;
@@ -7,7 +8,10 @@ use miniscript::bitcoin::{absolute, transaction, OutPoint, Psbt, Sequence};
 use miniscript::psbt::PsbtExt;
 use rand_core::RngCore;
 
-use crate::{apply_anti_fee_sniping, AntiFeeSnipingError, Finalizer, Input, InputMut, Output};
+use crate::{
+    apply_anti_fee_sniping, fisher_yates_shuffle, AntiFeeSnipingError, Finalizer, Input, InputMut,
+    Output,
+};
 
 /// Final selection of inputs and outputs.
 #[derive(Debug, Clone)]
@@ -171,6 +175,43 @@ impl Selection {
     /// coin-selection invariants — see [`InputMut`] for the available operations.
     pub fn inputs_mut(&mut self) -> impl Iterator<Item = InputMut<'_>> {
         self.inputs.iter_mut().map(InputMut::new)
+    }
+
+    /// Reorder inputs in-place using `compare`.
+    ///
+    /// Uses a stable sort: inputs that compare equal retain their relative order.
+    /// Typical use is BIP-69 lexicographic ordering by previous outpoint.
+    pub fn sort_inputs_by<F>(&mut self, compare: F)
+    where
+        F: FnMut(&Input, &Input) -> Ordering,
+    {
+        self.inputs.sort_by(compare);
+    }
+
+    /// Randomly shuffle inputs in-place using `rng`.
+    ///
+    /// Useful for chain-analysis resistance when no deterministic ordering is required.
+    pub fn shuffle_inputs<R: RngCore>(&mut self, rng: &mut R) {
+        fisher_yates_shuffle(&mut self.inputs, rng);
+    }
+
+    /// Reorder outputs in-place using `compare`.
+    ///
+    /// Uses a stable sort: outputs that compare equal retain their relative order.
+    /// Typical use is BIP-69 (ascending by amount, then by `script_pubkey`).
+    pub fn sort_outputs_by<F>(&mut self, compare: F)
+    where
+        F: FnMut(&Output, &Output) -> Ordering,
+    {
+        self.outputs.sort_by(compare);
+    }
+
+    /// Randomly shuffle outputs in-place using `rng`.
+    ///
+    /// Useful for chain-analysis resistance — in particular, hiding which output
+    /// is the change.
+    pub fn shuffle_outputs<R: RngCore>(&mut self, rng: &mut R) {
+        fisher_yates_shuffle(&mut self.outputs, rng);
     }
 
     /// Accumulates the maximum locktime from an iterator of input-required locktimes.
@@ -769,5 +810,14 @@ mod tests {
             matches!(result, Err(AntiFeeSnipingError::UnsupportedVersion(_))),
             "should return UnsupportedVersion error for version < 2"
         );
+    }
+
+    #[test]
+    fn test_fisher_yates_shuffle_preserves_multiset() {
+        let original: Vec<u32> = (0..32).collect();
+        let mut shuffled = original.clone();
+        fisher_yates_shuffle(&mut shuffled, &mut OsRng);
+        shuffled.sort();
+        assert_eq!(shuffled, original);
     }
 }
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -96,8 +96,6 @@ impl Default for PsbtParams {
 /// Occurs when creating a psbt fails.
 #[derive(Debug)]
 pub enum CreatePsbtError {
-    /// Attempted to mix locktime types.
-    LockTypeMismatch,
     /// Missing tx for legacy input.
     MissingFullTxForLegacyInput(Box<Input>),
     /// Missing tx for segwit v0 input.
@@ -119,7 +117,6 @@ impl From<AntiFeeSnipingError> for CreatePsbtError {
 impl core::fmt::Display for CreatePsbtError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            CreatePsbtError::LockTypeMismatch => write!(f, "cannot mix locktime units"),
             CreatePsbtError::MissingFullTxForLegacyInput(input) => write!(
                 f,
                 "legacy input that spends {} requires PSBT_IN_NON_WITNESS_UTXO",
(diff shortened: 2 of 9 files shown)


src/selector.rs
```diff
@@ -465,24 +490,20 @@ impl<'c> Selector<'c> {
...
+        let mut outputs = self.target_outputs.clone();
+        if maybe_change.is_some() {
+            outputs.push(Output::from((
+                self.change_script.clone(),
+                Amount::from_sat(maybe_change.value),
+            )));
+        }
+        Some(Selection::new(inputs, outputs))
     }
 }
```
noahjoeris, line 511: Should we add a test for mixed locktime units? Here's an example: 
```suggestion
}

#[cfg_attr(coverage_nightly, coverage(off))]
#[cfg(test)]
mod tests {
    use super::*;
    use crate::Input;
    use alloc::string::ToString;
    use bitcoin::{
        absolute,
        secp256k1::{Secp256k1, SecretKey},
        transaction, Amount, PrivateKey, Transaction, TxIn, TxOut,
    };
    use miniscript::{plan::Assets, Descriptor, DescriptorPublicKey};

    fn setup_cltv_input(cltv: absolute::LockTime) -> anyhow::Result<Input> {
        let secp = Secp256k1::new();
        let secret_key = SecretKey::from_slice(&[1_u8; 32])?;
        let public_key = PrivateKey::new(secret_key, bitcoin::Network::Regtest).public_key(&secp);
        let desc_str = format!("wsh(and_v(v:pk({public_key}),after({cltv})))");
        let desc_pk: DescriptorPublicKey = public_key.to_string().parse()?;
        let (desc, _) = Descriptor::parse_descriptor(&secp, &desc_str)?;
        let plan = desc
            .at_derivation_index(0)?
            .plan(&Assets::new().add(desc_pk).after(cltv))
            .expect("locktime asset must satisfy descriptor");
        let prev_tx = Transaction {
            version: transaction::Version::TWO,
            lock_time: absolute::LockTime::ZERO,
            input: vec![TxIn::default()],
            output: vec![TxOut {
                script_pubkey: desc.at_derivation_index(0)?.script_pubkey(),
                value: Amount::ONE_BTC,
            }],
        };
        Ok(Input::from_prev_tx(plan, prev_tx, 0, None)?)
    }

    #[test]
    fn test_selector_rejects_mixed_absolute_locktime_units() -> anyhow::Result<()> {
        let height_locked_input = setup_cltv_input(absolute::LockTime::from_consensus(10_000))?;
        let time_locked_input = setup_cltv_input(absolute::LockTime::from_consensus(500_000_001))?;
        let candidates = InputCandidates::new([], [height_locked_input, time_locked_input]);
        let params = SelectorParams::new(
            FeeRate::ZERO,
            vec![],
            ChangeScript::from_script(ScriptBuf::new(), Weight::ZERO),
        );

        assert!(matches!(
            Selector::new(&candidates, params),
            Err(SelectorError::LockTypeMismatch)
        ));

        Ok(())
    }
}
```

src/input.rs
```diff
@@ -665,6 +665,39 @@ impl Input {
...
+/// mutation to operations that preserve [`Selection`]'s coin-selection invariants.
+///
+/// Read-only access to the underlying [`Input`] is available via [`Deref`].
+///
+/// [`Selection`]: crate::Selection
+/// [`Selection::input_mut`]: crate::Selection::input_mut
+/// [`Selection::inputs_mut`]: crate::Selection::inputs_mut
+/// [`Deref`]: core::ops::Deref
+#[derive(Debug)]
+pub struct InputMut<'a>(&'a mut Input);
```
noahjoeris, line 680: Could we also simply use `Selection::set_input_sequence(...)`?

noahjoeris (commented):
Looking good overall. Left two comments

src/selector.rs
```diff
@@ -387,9 +397,24 @@ impl<'c> Selector<'c> {
...
+        let mut unit: Option<bitcoin::absolute::LockTime> = None;
+        for input in candidates.inputs() {
+            match (input.absolute_timelock(), unit) {
+                (Some(lt), Some(existing_unit)) if !existing_unit.is_same_unit(lt) => {
+                    return Err(SelectorError::LockTypeMismatch)
+                }
+                (Some(lt), None) => unit = Some(lt),
+                _ => {}
+            }
+        }
```
noahjoeris, line 418: nit: wdyt about this rewrite?
```suggestion
let mut unit_reference: Option<bitcoin::absolute::LockTime> = None;
for locktime in candidates
    .inputs()
    .filter_map(|input| input.absolute_timelock())
{
    let unit_reference = *unit_reference.get_or_insert(locktime);
    if !unit_reference.is_same_unit(locktime) {
        return Err(SelectorError::LockTypeMismatch);
    }
}
```

src/selector.rs
nymius: Yes!

src/selection.rs
```diff
@@ -279,6 +339,22 @@ impl Selection {
...
+        let j = loop {
+            let value = rng.next_u32();
+            if value >= threshold {
+                break (value % bound) as usize;
+            }
+        };
+        slice.swap(i, j);
+    }
+}
+
```
nymius: This is aiming for the same https://github.com/bitcoindevkit/bdk-tx/blob/8b51d0733565c14a1757636940ba46ba7de3b1ac/src/utils.rs#L162-L176

Could you compose and unify?

nymius (commented):
cACK 454ae11e005f0f26699eb378804d652fdcb4ddb0

I'm missing an example showing that `Selection` cannot be done because lock times doesn't match, which later splits the selection candidates and performs selection on one or both subsets successfully.

src/input.rs
evanlinjin: Exploring options...

```rust
impl Selection {
    /// I dislike the `Option<Result>>` signature.
    pub fn set_input_sequence(&mut self, op: OutPoint, seq: Sequence) -> Option<Result<(), Err>>;

    /// Hard for caller to discern between "not exist" vs "bad sequence".
    pub fn set_input_sequence(&mut self, op: OutPoint, seq: Sequence -> Result<(), ErrThatWrapsVeryAnnoying>;
}

// Option<Result<>> API:
selection.set_input_sequence(op, seq).expect("must exist")?;

// Original PR:
selection.input_mut(op).expect("must exist").set_sequence(seq)?;

// Wrapped error: Not worth exploring, very annoying.
```

The original PR API has the benefit of being able to iterate and mutate all inputs in one go. Additionally, if in the future we want to add more mutable methods on `Input`, we won't *pollute* the `Selection` method set (this is a weaker argument because YAGNI).

Personally, I think the `InputMut` API edges out a little bit more. What do you think?

src/selection.rs
evanlinjin: 3ca51f357e6ed64e10422d9eb54d4a34293b6e8b

src/selector.rs
evanlinjin: Looks good. I think `.filter_map` improves the clarity, I'll do that.

evanlinjin: Addressed, let me know what you think.

src/selector.rs
evanlinjin: Added.

src/input.rs
noahjoeris: Yeah I agree. The error wrapping sucks more. So let's keep `InputMut`

src/selector.rs
noahjoeris: Nice

noahjoeris (approved):
ACK 4b3a3e1

src/selection.rs
```diff
@@ -770,4 +811,13 @@ mod tests {
             "should return UnsupportedVersion error for version < 2"
         );
     }
+
+    #[test]
+    fn test_fisher_yates_shuffle_preserves_multiset() {
```
nymius, line 805: This should be in `no_std_rand.rs`

src/selection.rs
nymius, line 1: I would move this together with `afs.rs` rename and `no_std_rand.rs` to a new PR, and leave the breaking change alone in this one, as this is exclusively a new feature addition.

nymius (changes requested):
code ACK 4b3a3e1da8763bcfb49d77bdf0d56207f3366364

Without further changes to the code, I would split this into two PRs to avoid burying the input/output sorting changes below a breaking change.

One would contain the breaking commits and the tests (3 commits in total), the other the `afs.rs` rename and the `no_std_rand.rs` addition (including the `fisher_yates` test from `selection.rs`)

src/selection.rs
evanlinjin: Encapsulation is needed otherwise API hole is still there (caller can introduce conflicting locktime requires). Sort/shuffle needs to be added on top so users don't lose ordering capability between releases.

I initially preferred keeping `fisher_yates_shuffle` local, you pointed out the duplication, I agreed. Moving it out means it needs a new home. `no_std_rand.rs` is the rational choice. This was also your ask in #67.

I think it's best to keep all these changes in one PR.

evanlinjin:
> Without further changes to the code, I would split this into two PRs to avoid burying the input/output sorting changes below a breaking change.

I like the direction you are going with this policy, but it's too strict. I think it's good enough to make sure breaking changes are contained in commits that don't mix in non-breaking changes.

Sometimes you need to mix breaking/non-breaking changes in a single PR for the changes to make sense (like what is happening here).

src/selection.rs
nymius: Sorry, I didn't mean to move `fisher_yates`, what I mean was to apply the changes in a different PR.

nymius:
> Sometimes you need to mix breaking/non-breaking changes in a single PR for the changes to make sense (like what is happening here).

I agree, what I'm not seeing is the relation between the shuffle changes and the locktime ones. Could you expand yourself there?

evanlinjin:
> > Sometimes you need to mix breaking/non-breaking changes in a single PR for the changes to make sense (like what is happening here).
> 
> I agree, what I'm not seeing is the relation between the shuffle changes and the locktime ones. Could you expand yourself there?

Because we encapsulated `Selection`, you essentially cannot do anything with it, where as before, it had public fields so you can shuffle, rearrange directly.

nymius (approved):
ACK 96b9f883fcb0e6636c7dd50ba043e40d4eb2cf1b

noahjoeris (approved):
ACK 96b9f88

PR #73, opened 2026-05-20
feat: introduce TxTemplate as an intermediate stage

### LINKED ISSUE #57: Introduce `TxTemplate` as a state between `Selection` and `Psbt`
## Background

After a recent revisit of the codebase, I was dissatisfied with the API shape as it is now.

Initially, I blamed it on how AFS was added in the codebase. I left a post-merge review on [#5 here](https://github.com/bitcoindevkit/bdk-tx/pull/5#pullrequestreview-4243343806). The problems I pointed out can be summarized as:

* People who don't care about AFS now need to handle 2 additional error variants and an extra input.
* AFS is not "PSBT-creation-related". It's just an algorithm for deciding on locktime or input sequence values.
* There were some complex interactions between `PsbtInput` fields after the PR.

I suggested that AFS should be a step after `create_psbt`. @aagbotemi created a second PR to address my review [here (#56)](https://github.com/bitcoindevkit/bdk-tx/pull/56).

After looking at this PR, I realized my suggestions were incomplete. I did another thorough investigation and I've come up with a more complete direction/approach (which I will address here). Let's start with talking about AFS.

### Where should AFS go?

- **AFS should be a separate step to PSBT-creation**
    This is where my inital review of #5 got right. We don't want to pollute `.create_psbt` with inputs and error variants that are only relevant to AFS.

- **We should have the AFS step BEFORE psbt creation**
    This is my new realization. AFS is essentially just logic which decides on locktime/sequence. After the decision is made, this should be easily passed into creating a psbt. Additionally, this is how it's done in Bitcoin Core. [`DiscourageFeeSniping` is a step before PSBT creation](https://github.com/bitcoin/bitcoin/blob/10ca73c02cbff59f2134c0c7da3b8d0a7e727475/src/wallet/rpc/spend.cpp#L108).

However, since AFS logic currently depends on values contained in `PsbtParams`, it's difficult to implement it elsewhere. We need an architectural refactor.

## Proposal

I propose putting an additional stage in our transaction-creation pipeline called `TxTemplate`. This stage will sit between `Selection` and `Psbt`. Unlike the `Selection` type, `TxTemplate` will have complete information to create a `bitcoin::Transaction` (including the version, locktime and sequence fields which currently all reside in `PsbtParams`).

```rust
let mut rand = rand::thread_rng();

let mut (psbt, finalizer) = selection
    .into_tx_template(TxTemplateParams::default())
    .shuffle_inputs(&mut rand)
    .apply_anti_fee_sniping(tip_height, &mut rand)?
    .create_psbt(PsbtBuildParams::default())?;

let _ = psbt.sign(&signer, &secp);
let res = finalizer.finalize(&mut psbt);
assert!(res.is_finalized());
```

### Benefits of `TxTemplate`

* Clearer separation of concern. `TxTemplateParams` can purely be for deciding on bitcoin transaction fields. `PsbtParams` can be for PSBT-specific options.
* Callers that do not wish to work with `Psbt`s are no longer forced to. Additionally, when we add [support for PSBT V2](https://github.com/bitcoindevkit/bdk-tx/issues/48) in the future, we have a clean way to do it as a method on `TxTemplate::create_psbt_v2()`.
* AFS can be a single method on `TxTemplate`.

### Shape of `TxTemplate` and `TxTemplateParams`

```rust
/// Parameters for resolving a [`Selection`] into a [`TxTemplate`].
///
/// Carries only the *floors* and *defaults* that must be resolved at
/// template construction. Optional transformations (anti-fee-sniping,
/// shuffling, etc.) are exposed as methods on [`TxTemplate`].
#[derive(Debug, Clone)]
pub struct TxTemplateParams {
    /// Minimum tx version. Acts as a floor on `tx.version`.
    ///
    /// Default: [`transaction::Version::TWO`].
    pub min_version: transaction::Version,

    /// Minimum tx lock_time. Acts as a floor on `tx.lock_time`.
    ///
    /// Default: [`absolute::LockTime::ZERO`].
    pub min_locktime: absolute::LockTime,

    /// Default sequence used for plan-based inputs that don't specify their
    /// own (via plan-required relative timelock or via
    /// [`Input::set_sequence`]).
    ///
    /// Default: [`FALLBACK_SEQUENCE`] (Bitcoin Core's wallet default).
    pub fallback_sequence: Sequence,
}

/// A fully-resolved tx shape, intermediate between [`Selection`] and the
/// final [`Psbt`] / [`bitcoin::Transaction`].
#[derive(Debug, Clone)]
pub struct TxTemplate {
    pub(crate) version: transaction::Version,
    pub(crate) lock_time: absolute::LockTime,
    pub(crate) inputs: Vec<Input>,
    pub(crate) outputs: Vec<Output>,
}
```

In order for the above shape to work, the `Input` type should be able to specify the sequence and sighash values.

* #60 
* #61

evanlinjin:
### Description

Closes #57. Based on #87.

`PsbtParams` was doing two jobs: it carried bitcoin-transaction-shape fields (`version`, `min_locktime`, sequence) *and* PSBT emission options. Anti-fee-sniping depended on the tx-shape half, so it was structurally pinned inside `create_psbt` — which meant every caller reckoned with two AFS error variants and an extra input whether they wanted AFS or not.

AFS isn't a PSBT concern. It decides a transaction's locktime and input sequences, so it belongs *before* emission — as in Bitcoin Core, where [`DiscourageFeeSniping` runs as a step before the transaction is built](https://github.com/bitcoin/bitcoin/blob/10ca73c02cbff59f2134c0c7da3b8d0a7e727475/src/wallet/rpc/spend.cpp#L108). It couldn't move without first splitting the tx-shape fields out of `PsbtParams`.

This PR renames `Selection` to `TxTemplate` and gives it those fields, so the tx shape is owned and observable before anything is emitted:

```
Selector::try_finalize() → TxTemplate → SealedTxTemplate → (Psbt, Finalizer) | Transaction
```

Mostly ergonomics, with one exception. On `master`, `PsbtParams { version: ONE, .. }` with a relative-timelock input builds a v1 transaction whose `OP_CSV` can never be satisfied — BIP-68 sequence locks only apply from v2 — and you find out when it's rejected. That's now `SetVersionError::RelativeTimelockRequiresV2`.

### Notes to the reviewers

Commit 1 is a pure rename — `TxTemplate` *is* `Selection`, not a new layer between it and the PSBT.

Three commits at the end respond to review:

- `discourage_fee_sniping` (was `apply_anti_fee_sniping`). `apply_*` reads as an independent transformation that composes with the other shaping methods, but AFS shares the `lock_time` slot with `set_locktime`. Also matches Core's naming.
- A docs commit stating how those two compose. `set_locktime` is a *floor*: `tx.lock_time` only moves up from there, AFS may raise it toward the tip but never lowers it — the same way input CLTVs already compose. I considered making AFS authoritative instead (last-write-wins) and decided against it: having AFS silently discard a value the caller explicitly set is the footgun this PR exists to remove, and it would make AFS the one operation in `TxTemplate` that can *lower* `lock_time`.
- AFS is sealed: `discourage_fee_sniping` consumes the template and returns a `SealedTxTemplate` exposing only reads and emission. This is narrower than it sounds — only AFS seals, so `set_locktime` before it still works. It rules out one concrete case: AFS may protect the transaction by setting an input's `nSequence` rather than the locktime, which it only does while `lock_time` is zero, and a `set_locktime` afterwards would rebuild a transaction carrying both a near-tip locktime and a confirmation-depth sequence — exactly the fingerprint #87 prevents. A `compile_fail` doctest pins it.

Points 1 and 3 of https://github.com/bitcoindevkit/bdk-tx/pull/73#issuecomment-5297469748 are not addressed here; point 2 is #87, which this builds on.

### Changelog notice

- Renamed `Selection` to `TxTemplate`, now the single workspace for transaction shaping (version, locktime, fallback sequence, per-input sequence, ordering, anti-fee-sniping, emission).
- `Selector::try_finalize` now returns `Option<TxTemplate>`; `InputCandidates::into_selection` is now `into_tx_template`, returning `Result<TxTemplate, IntoTxTemplateError>` (was `IntoSelectionError`).
- Split `PsbtParams` into `BuildPsbtParams` (emission-only). Tx-shape options moved to `TxTemplate` setters: `set_version`, `set_locktime`, `set_fallback_sequence`, `discourage_fee_sniping`.
- PSBT emission is `build_psbt` (renamed from `create_psbt`) and returns `(Psbt, Finalizer)`; its params/error are `BuildPsbtParams` / `BuildPsbtError`. `shuffle_inputs` is now consuming (returns `Self`).
- Anti-fee-sniping is `TxTemplate::discourage_fee_sniping` (was the `PsbtParams::anti_fee_sniping` field). It and `set_locktime` write the same `tx.lock_time` slot and compose monotonically: a value passed to `set_locktime` is a floor that AFS may raise toward the chain tip but never lowers.
- Anti-fee-sniping is terminal: `discourage_fee_sniping` consumes the `TxTemplate` and returns a new `SealedTxTemplate` that exposes only reads and emission, so the tx shape cannot be mutated after AFS. `TxTemplate` derefs to `SealedTxTemplate`.
- Behaviour change: previously-silent locktime handling is now explicit. A wrong-unit `min_locktime` was silently ignored and a below-CLTV value silently clamped; `set_locktime` now returns `SetLockTimeError::UnitMismatch` / `BelowInputCltv`, and `set_version` returns `SetVersionError::RelativeTimelockRequiresV2`. The hardcoded `ENABLE_RBF_NO_LOCKTIME` fallback sequence is now configurable via `set_fallback_sequence` (default unchanged).

### Before submitting

### DIFF
--- a/examples/synopsis.rs
+++ b/examples/synopsis.rs
@@ -52,7 +52,7 @@ fn main() -> anyhow::Result<()> {
         .all_candidates()
         .regroup(group_by_spk())
         .filter(filter_unspendable(tip_height, Some(tip_mtp)))
-        .into_selection(
+        .into_tx_template(
             selection_algorithm_lowest_fee_bnb(longterm_feerate, 100_000),
             SelectorParams {
                 // For waste-optimization when deciding change.
@@ -122,7 +122,7 @@ fn main() -> anyhow::Result<()> {
 
         let selection = rbf_candidates
             // Do coin selection.
-            .into_selection(
+            .into_tx_template(
                 // Coin selection algorithm.
                 selection_algorithm_lowest_fee_bnb(longterm_feerate, 100_000),
                 SelectorParams {
--- a/examples/synopsis.rs
+++ b/examples/synopsis.rs
@@ -1,6 +1,6 @@
 use bdk_testenv::{bitcoincore_rpc::RpcApi, TestEnv};
 use bdk_tx::{
-    filter_unspendable, group_by_spk, selection_algorithm_lowest_fee_bnb, Output, PsbtParams,
+    filter_unspendable, group_by_spk, selection_algorithm_lowest_fee_bnb, Output, PsbtBuildParams,
     SelectorParams, Signer,
 };
 use bitcoin::{key::Secp256k1, Amount, FeeRate};
@@ -48,7 +48,7 @@ fn main() -> anyhow::Result<()> {
         .assume_checked();
 
     // Okay now create tx.
-    let selection = wallet
+    let (mut psbt, finalizer) = wallet
         .all_candidates()
         .regroup(group_by_spk())
         .filter(filter_unspendable(tip_height, Some(tip_mtp)))
@@ -66,10 +66,8 @@ fn main() -> anyhow::Result<()> {
                     bdk_tx::ChangeScript::from_descriptor(internal.at_derivation_index(0)?),
                 )
             },
-        )?;
-
-    let mut psbt = selection.create_psbt(PsbtParams::default())?;
-    let finalizer = selection.into_finalizer();
+        )?
+        .create_psbt(PsbtBuildParams::default())?;
 
     let _ = psbt.sign(&signer, &secp);
     let res = finalizer.finalize(&mut psbt);
@@ -146,7 +144,6 @@ fn main() -> anyhow::Result<()> {
                 },
             )?;
 
-        let mut psbt = selection.create_psbt(PsbtParams::default())?;
         println!(
             "selected inputs: {:?}",
             selection
@@ -156,7 +153,7 @@ fn main() -> anyhow::Result<()> {
                 .collect::<Vec<_>>()
         );
 
-        let finalizer = selection.into_finalizer();
+        let (mut psbt, finalizer) = selection.create_psbt(PsbtBuildParams::default())?;
         psbt.sign(&signer, &secp).expect("failed to sign");
         assert!(
             finalizer.finalize(&mut psbt).is_finalized(),
--- a/examples/synopsis.rs
+++ b/examples/synopsis.rs
@@ -1,6 +1,6 @@
 use bdk_testenv::{bitcoincore_rpc::RpcApi, TestEnv};
 use bdk_tx::{
-    filter_unspendable, group_by_spk, selection_algorithm_lowest_fee_bnb, Output, PsbtBuildParams,
+    filter_unspendable, group_by_spk, selection_algorithm_lowest_fee_bnb, BuildPsbtParams, Output,
     SelectorParams, Signer,
 };
 use bitcoin::{key::Secp256k1, Amount, FeeRate};
@@ -67,7 +67,7 @@ fn main() -> anyhow::Result<()> {
                 )
             },
         )?
-        .create_psbt(PsbtBuildParams::default())?;
+        .build_psbt(BuildPsbtParams::default())?;
 
     let _ = psbt.sign(&signer, &secp);
     let res = finalizer.finalize(&mut psbt);
@@ -153,7 +153,7 @@ fn main() -> anyhow::Result<()> {
                 .collect::<Vec<_>>()
         );
 
-        let (mut psbt, finalizer) = selection.create_psbt(PsbtBuildParams::default())?;
+        let (mut psbt, finalizer) = selection.build_psbt(BuildPsbtParams::default())?;
         psbt.sign(&signer, &secp).expect("failed to sign");
         assert!(
             finalizer.finalize(&mut psbt).is_finalized(),
--- a/src/input_candidates.rs
+++ b/src/input_candidates.rs
@@ -7,8 +7,8 @@ use miniscript::bitcoin;
 
 use crate::collections::{BTreeMap, HashSet};
 use crate::{
-    CannotMeetTarget, FeeRateExt, Input, InputGroup, Selection, Selector, SelectorError,
-    SelectorParams,
+    CannotMeetTarget, FeeRateExt, Input, InputGroup, Selector, SelectorError, SelectorParams,
+    TxTemplate,
 };
 
 /// Input candidates.
@@ -191,30 +191,30 @@ impl InputCandidates {
         self
     }
 
-    /// Attempt to convert the input candidates into a valid [`Selection`] with a given
+    /// Attempt to convert the input candidates into a valid [`TxTemplate`] with a given
     /// `algorithm` and selector `params`.
-    pub fn into_selection<A, E>(
+    pub fn into_tx_template<A, E>(
         self,
         algorithm: A,
         params: SelectorParams,
-    ) -> Result<Selection, IntoSelectionError<E>>
+    ) -> Result<TxTemplate, IntoTxTemplateError<E>>
     where
         A: FnMut(&mut Selector) -> Result<(), E>,
     {
-        let mut selector = Selector::new(&self, params).map_err(IntoSelectionError::Selector)?;
+        let mut selector = Selector::new(&self, params).map_err(IntoTxTemplateError::Selector)?;
         selector
             .select_with_algorithm(algorithm)
-            .map_err(IntoSelectionError::SelectionAlgorithm)?;
+            .map_err(IntoTxTemplateError::SelectionAlgorithm)?;
         let selection = selector
             .try_finalize()
-            .ok_or(IntoSelectionError::CannotMeetTarget(CannotMeetTarget))?;
+            .ok_or(IntoTxTemplateError::CannotMeetTarget(CannotMeetTarget))?;
         Ok(selection)
     }
 }
 
 /// Occurs when we cannot find a solution for selection.
 #[derive(Debug)]
-pub enum IntoSelectionError<E> {
+pub enum IntoTxTemplateError<E> {
     /// Coin selector returned an error
     Selector(SelectorError),
     /// Selection algorithm failed.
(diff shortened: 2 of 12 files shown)


evanlinjin:
This is fully AI generated and not ready for review.

examples/synopsis.rs
```diff
@@ -66,10 +66,8 @@ fn main() -> anyhow::Result<()> {
                     bdk_tx::ChangeScript::from_descriptor(internal.at_derivation_index(0)?),
                 )
             },
-        )?;
-
-    let mut psbt = selection.create_psbt(PsbtParams::default())?;
-    let finalizer = selection.into_finalizer();
+        )?
+        .build_psbt(BuildPsbtParams::default())?;
```
noahjoeris, line 70: Is it worth passing in `BuildPsbtParams::default` every time for one parameter? Do you think there will be more params in the future? What about having `build_psbt()` and `build_psbt_with(psbt_params)`?

src/tx_template.rs
```diff
@@ -0,0 +1,944 @@
...
+impl core::ops::Deref for TxTemplate {
+    type Target = SealedTxTemplate;
+
+    fn deref(&self) -> &SealedTxTemplate {
+        &self.0
+    }
+}
+
+impl TxTemplate {
+    pub(crate) fn from_parts(inputs: Vec<Input>, outputs: Vec<Output>) -> Self {
```
noahjoeris: nit: What about improving this name? Maybe `new` or `from_inputs_outputs`

noahjoeris (commented):
cACK [0f1c8b2](https://github.com/bitcoindevkit/bdk-tx/commit/0f1c8b27568f117b9d13b5bd908a64afc5f2cc62)

Nice improvements, thanks.

src/input_candidates.rs
```diff
@@ -191,30 +192,59 @@ impl InputCandidates {
...
-    /// Attempt to convert the input candidates into a valid [`Selection`] with a given
-    /// `algorithm` and selector `params`.
-    pub fn into_selection<A, E>(
+    /// Add `input` to the must-select group (always included by coin selection).
+    ///
+    /// All must-select inputs form a single group spent together. If `input`'s outpoint is already
+    /// a candidate, it is ignored and the existing candidate is kept.
+    pub fn push_must_select(mut self, input: impl Into<Input>) -> Self {
+        let input = input.into();
+        if self.contains.insert(input.prev_outpoint()) {
```
aagbotemi: Can the `push_can_select` and `push_must_select` be called on the same input in sequence, and if so, should the second call move the input into the must-select group? Right now it silently does nothing and the input stays in can-select.

examples/synopsis.rs
evanlinjin: Personally, I prefer having a single method as more methods means more documentation to maintain.

src/input_candidates.rs
evanlinjin: Good point. Changed to upset logic here: 0e7d0635c51b9177dd530717eb97e9a2833df23b.

Let me know what you think.

evanlinjin: Removed these two methods from this PR as they are out of scope.

ValuedMammal:
I'm not sold on the idea of TxTemplate I think it adds a layer of indirection with little benefit.

evanlinjin:
@ValuedMammal `TxTemplate` is `Selection` repurposed. Commit 1 is that pure-rename commit. There is no added indirection.

Could you be a bit more specific on where you think the proposed API is problematic? Which costs are you weighing?

nymius:
I agree with the merits of this change. It detaches AFS, version, locktime and sequence from the PSBT creation, and TxTemplate ensures the invariants for a single party.
For multi party setups we can make `SealedTxTemplate` check the invariants of its own subset of inputs/outputs for a Psbt too  (like `SealedTxTemplate::check_psbt`). In this case it would be nice to have `SealedTxTemplate` implement serialization/deserialization to match `Psbt`s with their original templates.
I've only been considering this from a conceptual level. Before proceeding with more review, I would like to be sure there are not blocking concerns about this. @ValuedMammal ?

ValuedMammal:
> @ValuedMammal `TxTemplate` is `Selection` repurposed. Commit 1 is that pure-rename commit. There is no added indirection.

Oh, the impression that TxTemplate is an indirection may have come from it being proposed as an intermediate stage; I can see that TxTemplate replaced the Selection type. The issue cites the anti-fee-sniping logic as an awkward API, but so far the shape of the API hasn't prevented me from creating PSBTs, so I don't know if I'm missing something.

I'm also not satisfied with the AFS logic, but for different reasons

1. Exclusion of p2tr inputs with a CSV condition https://github.com/bitcoindevkit/bdk-tx/pull/65#discussion_r3786494308
2. Non-zero LockTime in the Sequence branch https://github.com/bitcoindevkit/bdk-tx/pull/65#discussion_r3786557326
3. Unintuitive min_locktime interaction https://github.com/bitcoindevkit/bdk-tx/pull/65#discussion_r3786635906

evanlinjin:
> so far the shape of the API hasn't prevented me from creating PSBTs

@ValuedMammal You're right that the API in `master` does not stop you from creating PSBTs. This PR is not claiming to fix broken functionality. It is about improving ergonomics.

The new `TxTemplate` API provides a lot more transparency by returning the result and error at the call site. This also means the error types returned have fewer variants and would always be relevant to the caller. It is also easier to observe the resolved tx shape -- you can argue that the caller can build the PSBT and read it back out, but I would say that is awkward and it's not clear which parameter in `PsbtParams` did what to the final result.

Also mentioned already in the PR description (kinda):

* `rand` input is only applicable to AFS so it's better to not contaminate calls which don't intend on doing AFS with the extra param. The `TxTemplate` architecture makes it more transparent where and how it's used.
* Conceptually, it's cleaner to decouple AFS from PSBT-creation as AFS is not a PSBT concept.

In terms of the three points you raised:

1. This seems to be a nice-to-have, not a bug. Not relevant to this PR. Would you like me to create a ticket?
2. This is a serious bug and thanks for identifying it. Since it seems urgent, I've created a PR for it here: https://github.com/bitcoindevkit/bdk-tx/pull/87 and rebased this PR on top.
3. This is relevant to the PR. ~~The behavior I employed for `TxTemplate` is last-write-wins between `set_locktime` and `discourage_fee_sniping`: 04c5fc9188a7ce75fc9d4a0005b781913877314d.~~ The other solution would be to have `TxTemplate::set_locktime_policy` which takes in a `LocktimePolicy` (like you suggested), but with rand as part of the AFS variant.

    **Edit:** I decided to keep the PR's original behavior because a monotonically increasing `locktime` is predictable and expected. `locktime` will compose like input CLTVs which is easy to reason about and the caller's setting is never silently discarded.

PR #79, opened 2026-06-13
finalizer: BIP174 compliance + sighash/signature-size validation

evanlinjin:
### Description

Hardens `Finalizer` to follow BIP174 more closely and refuse to emit a transaction that would be malformed or unsafe to broadcast. The finalizer now validates signatures before assembling the final witness/scriptSig, and reports failures through a dedicated `FinalizeError` instead of leaking `miniscript::Error`.

More context here: https://github.com/bitcoindevkit/bdk-tx/pull/74#issuecomment-4587522648

What changed:

- **Retain unknown/proprietary input fields.** Per BIP174 a finalizer must clear all input data *except* the UTXO and unknown/proprietary fields; finalization now preserves `unknown` and `proprietary`.
- **`SighashMismatch`** — a signature's sighash type disagrees with the input's declared `PSBT_IN_SIGHASH_TYPE`. **Mandated by BIP174.**
- **`SighashNotAllowed`** — no `PSBT_IN_SIGHASH_TYPE` is declared, yet a signature uses neither `DEFAULT` nor `ALL`. Stricter-than-spec safeguard against silently changing signing semantics.
- **`SignatureTooLarge`** — a satisfied schnorr witness is larger than the size the spending `Plan` committed to (e.g. a 65-byte `SIGHASH_ALL` sig where 64-byte `SIGHASH_DEFAULT` was planned). Because the plan's weight/feerate estimate derives from the committed size, a heavier witness makes the tx undershoot its target feerate and risk being unbroadcastable. Only the *larger* direction is rejected; a smaller witness merely overpays and is allowed.
- **Dedicated `FinalizeError`** — `finalize_input` / `FinalizeMap` now return `FinalizeError`, modelling the failures above plus `Satisfaction(miniscript::Error)`.

### Notes to the reviewers

`SignatureTooLarge` is intentionally **not** unit-tested: it is only reachable in release builds. In debug, miniscript's `satisfy_self` hits a `debug_assert!` on the signature size and panics before our check runs - an upstream bug.

Out of scope, for follow-up:

- ECDSA signatures are estimated at their 73-byte max, so they can't exceed the plan and need no check today (`TODO` in code). An upstream change could give ECDSA placeholders a committed size to capture low-R grinding fee savings safely.
- The upstream rust-miniscript `debug_assert` → recoverable-error fix.

### Changelog notice

```md
Changed:
- **(breaking)** `Finalizer::finalize_input`, `FinalizeMap`, and `FinalizeMap::results` now return the new `FinalizeError` type instead of `miniscript::Error`.

Added:
- `FinalizeError` enum with `SighashMismatch`, `SighashNotAllowed`, `SignatureTooLarge`, and `Satisfaction` variants.
- Finalization now validates that each signature's sighash type respects `PSBT_IN_SIGHASH_TYPE` (or is `DEFAULT`/`ALL` when unset) and that satisfied witness sizes do not exceed the spending plan.

Fixed:
- Finalization now retains `unknown` and `PSBT_IN_PROPRIETARY` input fields, as required by BIP174.
```

### Before submitting

### DIFF
--- a/src/finalizer.rs
+++ b/src/finalizer.rs
@@ -303,9 +303,9 @@ impl core::error::Error for FinalizeError {}
 #[cfg_attr(coverage_nightly, coverage(off))]
 #[cfg(test)]
 mod tests {
-    use crate::{Finalizer, Output, PsbtParams, Selection, Signer};
+    use crate::{FinalizeError, Finalizer, Output, PsbtParams, Selection, Signer};
     use bitcoin::secp256k1::Secp256k1;
-    use bitcoin::{absolute, transaction, Amount, ScriptBuf, TxIn, TxOut};
+    use bitcoin::{absolute, transaction, Amount, ScriptBuf, TapSighashType, TxIn, TxOut};
     use miniscript::bitcoin;
     use miniscript::bitcoin::Transaction;
     use miniscript::plan::Assets;
@@ -553,6 +553,50 @@ mod tests {
         assert!(results.is_empty());
         assert_eq!(psbt.inputs[0].final_script_sig, final_script_sig);
         assert_eq!(psbt.inputs[0].final_script_witness, final_script_witness);
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_finalize_sighash_mismatch() -> anyhow::Result<()> {
+        let (input, keymap) = create_input_from_descriptor_at(TR_XPRV, 0)?;
+        let output = Output::with_script(ScriptBuf::new(), Amount::from_sat(9_000));
+        let selection = Selection::new(vec![input], vec![output]);
+
+        let mut psbt = selection.create_psbt(PsbtParams::default())?;
+        let finalizer = selection.into_finalizer();
+        psbt.sign(&Signer(keymap), &Secp256k1::new())
+            .expect("signing failed");
+
+        // The signature commits to DEFAULT, but we declare ALL, so the two disagree.
+        psbt.inputs[0].sighash_type = Some(TapSighashType::All.into());
+
+        let err = finalizer.finalize_input(&mut psbt, 0).unwrap_err();
+        assert!(matches!(err, FinalizeError::SighashMismatch { .. }));
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_finalize_sighash_not_allowed() -> anyhow::Result<()> {
+        let (input, keymap) = create_input_from_descriptor_at(TR_XPRV, 0)?;
+        let output = Output::with_script(ScriptBuf::new(), Amount::from_sat(9_000));
+        let selection = Selection::new(vec![input], vec![output]);
+
+        let mut psbt = selection.create_psbt(PsbtParams::default())?;
+        let finalizer = selection.into_finalizer();
+        psbt.sign(&Signer(keymap), &Secp256k1::new())
+            .expect("signing failed");
+
+        // No PSBT_IN_SIGHASH_TYPE declared, yet the signature uses neither DEFAULT nor ALL.
+        psbt.inputs[0]
+            .tap_key_sig
+            .as_mut()
+            .expect("tap key sig")
+            .sighash_type = TapSighashType::Single;
+
+        let err = finalizer.finalize_input(&mut psbt, 0).unwrap_err();
+        assert!(matches!(err, FinalizeError::SighashNotAllowed { .. }));
 
         Ok(())
     }
--- a/src/finalizer.rs
+++ b/src/finalizer.rs
@@ -1,6 +1,6 @@
 use crate::collections::{BTreeMap, HashMap};
-use bitcoin::{OutPoint, Psbt, Witness};
-use miniscript::{bitcoin, plan::Plan, psbt::PsbtInputSatisfier};
+use bitcoin::{psbt::PsbtSighashType, OutPoint, Psbt, Witness};
+use miniscript::{bitcoin, miniscript::satisfy::Placeholder, plan::Plan, psbt::PsbtInputSatisfier};
 
 /// Type used to finalize inputs of a Partially Signed Bitcoin Transaction (PSBT) using
 /// a collection of pre-computed spending plans.
@@ -11,6 +11,9 @@ use miniscript::{bitcoin, plan::Plan, psbt::PsbtInputSatisfier};
 /// the pre-computed spending [`Plan`] for each input. This process converts a PSBT input from a
 /// partially signed state to a fully signed state, making it ready for extraction into a valid
 /// Bitcoin [`Transaction`].
+///
+/// This type fills the [BIP174] *Input Finalizer* role: it consumes signatures already present in
+/// the PSBT and assembles the final witness/scriptSig.
 ///
 /// # Usage
 ///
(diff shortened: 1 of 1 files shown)


src/finalizer.rs
```diff
@@ -58,68 +61,136 @@ pub struct Finalizer {
...
+        // If unset, only permit ALL/DEFAULT (stricter-than-spec safeguard).
+        let mut psbt_in_sighashes = {
+            let partial_sigs = psbt_in.partial_sigs.values().map(|s| s.sighash_type as u32);
+            let tap_key_sig = psbt_in.tap_key_sig.iter().map(|s| s.sighash_type as u32);
+            let tap_script_sigs = psbt_in
+                .tap_script_sigs
+                .values()
+                .map(|s| s.sighash_type as u32);
+            partial_sigs.chain(tap_key_sig).chain(tap_script_sigs)
+        };
```
nymius, line 134: Why would an input have `partial_sigs` and `tap_script_sigs`? Shouldn't that be an error? I think we should check the sighashes in an "ad hoc" fashion, for each kind of signature, and avoid reducing them to the same thing.

src/finalizer.rs
```diff
@@ -58,68 +61,136 @@ pub struct Finalizer {
...
+            let exp_sighash_type = in_sighash_type.to_u32();
+            if let Some(sighash_mismatch) = psbt_in_sighashes.find(|&t| t != exp_sighash_type) {
+                return Err(FinalizeError::SighashMismatch {
+                    expected: PsbtSighashType::from_u32(exp_sighash_type),
+                    got: PsbtSighashType::from_u32(sighash_mismatch),
+                });
             }
-            if !stack.is_empty() {
-                psbt_input.final_script_witness = Some(Witness::from_slice(&stack));
+        } else if let Some(sighash_mismatch) = psbt_in_sighashes.find(|&t| t > 0x01 /*ALL*/) {
```
nymius, line 143: There is a `PsbtSighashFlag::ALL` for 0x01.

src/finalizer.rs
```diff
@@ -58,68 +61,136 @@ pub struct Finalizer {
...
+        //
+        // Only schnorr placeholders are checked, because schnorr is the only signature type whose
+        // size is a plan-time choice: 64 bytes for SIGHASH_DEFAULT vs 65 for an explicit sighash.
+        //
+        // TODO: Add ECDSA checks once upstream adds them.
+        for (temp, stack_item) in plan.witness_template().iter().zip(&stack) {
+            if let Placeholder::SchnorrSigPk(_, _, size)
+            | Placeholder::SchnorrSigPkHash(_, _, size) = temp
+            {
+                // Only a witness *larger* than the plan is dangerous.
```
nymius, line 163: How should a recovery mechanism look like in the case we want to save the extra sats? Should we return an enum here with "Finalized, NotFinalized, Overshoot" or something similar? Should we compare the target fee rate and inner fee rate at the end of the tx creation process and rewind if they don't match?

src/finalizer.rs
```diff
@@ -157,17 +228,84 @@ impl FinalizeMap {
...
+impl core::fmt::Display for FinalizeError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            FinalizeError::SighashMismatch { expected, got } => write!(
+                f,
+                "signature has sighash type ({got}) when ({expected}) is declared in PSBT_IN_SIGHASH_TYPE"
+            ),
+            FinalizeError::SighashNotAllowed { got } => write!(
+                f,
+                "signature has sighash type ({got}) but no PSBT_IN_SIGHASH_TYPE is declared; only ALL or DEFAULT are permitted"
```
nymius: This message is confusing, `... only ALL or DEFAULT are permitted`: this is always or only when `PSBT_IN_SIGHASH_TYPE` isn't declared?

nymius (commented):
cACK 0291c170caa088e1d2a1743da3671f8ba5946b21

src/finalizer.rs
evanlinjin: I don't think the finalizer should reject every invalid state a PSBT can represent. It's job is to produce a correct final witness.

> I think we should check the sighashes in an "ad hoc" fasihon, for each kind of signature, and avoid reducing them to the same thing.

A per-kind check would literally do the exact same thing though.

src/finalizer.rs
evanlinjin: That type doesn't exist in the version of `bitcoin`/`miniscript` we are using here.

src/finalizer.rs
evanlinjin: Interesting question. I think we should reserve this PR purely for safety. Everything else can be tackled in separate PRs. Could you create a ticket? Thanks!

src/finalizer.rs
evanlinjin: Fixed in 71294863aff3616f4dc12b0178b9af9d2887f1a1

src/finalizer.rs
nymius: But would remove this question `Why would an input have partial_sigs and tap_script_sigs?` altogether at the same time, and make clear that we are not considering them the same thing.

src/finalizer.rs
```diff
@@ -58,68 +61,136 @@ pub struct Finalizer {
...
+    /// unknown/proprietary fields are retained.
     ///
     /// # Errors
     ///
-    /// If the spending plan associated with the PSBT input cannot be satisfied,
-    /// then a [`miniscript::Error`] is returned.
+    /// Returns a [`FinalizeError`]:
+    ///
+    /// * [`SighashMismatch`] - a signature's sighash type disagrees with the input's declared
+    ///   `PSBT_IN_SIGHASH_TYPE`.
```
nymius, line 87:
```suggestion
    /// * [`SighashMismatch`] - a signature's sighash type disagrees with the input's declared in
    ///   `PSBT_IN_SIGHASH_TYPE`.
```

nymius (commented):
This is the only way I found to reproduce the `SignatureTooLarge` error:

```rust
    #[test]
    #[cfg(not(debug_assertions))]
    fn test_finalize_signature_too_large() -> anyhow::Result<()> {
        let (non_default_input, input, keymap) = {
            let secp = Secp256k1::new();
            let (desc, keymap) = Descriptor::parse_descriptor(&secp, TR_XPRV)?;
            let def_desc = desc.at_derivation_index(0)?;
            let script_pubkey = def_desc.script_pubkey();

            let assets = keymap.keys().fold(Assets::new(), |a, k| a.add(k.clone()));
            let mut non_default_assets = keymap.keys().fold(Assets::new(), |a, k| a.add(k.clone()));
            non_default_assets.keys = std::mem::take(&mut non_default_assets.keys)
                .into_iter()
                .map(|(key_source, mut can_sign)| {
                    can_sign.taproot.sighash_default = false;
                    (key_source, can_sign)
                })
                .collect();

            let plan = def_desc.clone().plan(&assets).expect("failed to create plan");
            let non_default_plan = def_desc.plan(&non_default_assets).expect("failed to create plan");

            let prev_tx = Transaction {
                version: transaction::Version::TWO,
                lock_time: absolute::LockTime::ZERO,
                input: vec![TxIn::default()],
                output: vec![TxOut {
                    script_pubkey,
                    value: Amount::from_sat(100_000),
                }],
            };

            let status = crate::ConfirmationStatus::new(1_000, Some(500_000_000))?;
            let non_default_input = crate::Input::from_prev_tx(non_default_plan, prev_tx.clone(), 0, Some(status))?;
            let input = crate::Input::from_prev_tx(plan, prev_tx, 0, Some(status))?;
            (non_default_input, input, keymap)
        };
        let output = Output::with_script(ScriptBuf::new(), Amount::from_sat(9_000));
        let non_default_selection = Selection::new(vec![non_default_input], vec![output.clone()]);

        let mut psbt = non_default_selection.create_psbt(PsbtParams::default())?;

        psbt.inputs[0].sighash_type = Some(TapSighashType::All.into());
        psbt.sign(&Signer(keymap), &Secp256k1::new())
            .expect("signing failed");

        psbt.inputs[0].sighash_type = None;

        let selection = Selection::new(vec![input], vec![output]);
        let finalizer = selection.into_finalizer();
        let err = finalizer.finalize_input(&mut psbt, 0).unwrap_err();
        assert!(matches!(err, FinalizeError::SignatureTooLarge { .. }));

        Ok(())
    }
```

I run it with:
```bash
cargo test --release -- test_finalize_signature_too_large
```

because there is a `debug_assert` in the satisfier that was panicking earlier and preventing the error to happen.

src/finalizer.rs
nymius: Line 137 prevents any different sighash type to happen, so as long as it is in place, there's no way a partial sig in a taproot signed input affects the behavior of this method.

PR #82, opened 2026-06-27
refactor!: Replace `Selector` with a single `into_selection` pass

evanlinjin:
### Description

Reworks the coin-selection API around a single `InputCandidates::into_selection` pass and cleans up its error surface.

The public `Selector` struct is removed as nothing used it. Its methods were only ever called in sequence by `into_selection`, so it was public surface with no consumer. That logic now lives inside `into_selection`, and algorithms take a `bdk_coin_select::CoinSelector` plus a pure-data `SelectionContext` instead of a `&mut Selector`.

This also:
- unifies the two redundant long-term feerates (change policy + bnb metric) into one `SelectionParams::longterm_feerate`;
- splits the catch-all `SelectorError` so each error type only carries variants it can actually produce;
- fixes a reachability bug that produced a false `CannotMeetTarget` under RBF.

### Notes to the reviewers

- **Behavior change (please scrutinize):** the change policy is now *always* waste-aware (`min_value_and_waste`). Previously a `None` long-term feerate selected a plain dust-only policy. Consequence: raw-script change must now supply a real `satisfaction_weight` (descriptor change is unaffected — it's derived). Rationale: `min_value_and_waste` with `longterm == target` does not collapse to dust-only; it still charges the change output's lifetime cost, which is the correct thing to weigh.
- **Bug fix detail:** the reachability pre-check called `select_all_effective(raw target_feerate)` but scored with `excess(target.fee.rate)` (the RBF-resolved, higher rate). This could result in an under-selection. Fixed by selecting at `target.fee.rate`.
- **Error split:** `to_cs_change_policy` previously returned a `SelectorError` that included a `LockTypeMismatch` it could never produce. It now returns `ChangePolicyError` (only `Miniscript` / `InsufficientAssets`), and `LockTypeMismatch` is a top-level `IntoSelectionError` variant.

### Changelog notice

```md
Added:
- Add `SelectionContext`, passed to selection algorithms alongside the `CoinSelector`.
- Add `ChangePolicyError` (`Miniscript`, `InsufficientAssets`) as the return type of `SelectionParams::to_cs_change_policy`.
- Add `IntoSelectionError` variants `ChangePolicy`, `LockTypeMismatch`, `CannotMeetTarget { missing }`, and `AlgorithmFellShort`.

Changed:
- Rename `SelectorParams` to `SelectionParams`.
- `InputCandidates::into_selection` no longer takes a `&mut Selector` closure; algorithms are now `FnOnce(&mut bdk_coin_select::CoinSelector, SelectionContext) -> Result<(), E>`.
- Replace `SelectionParams::change_longterm_feerate` with `SelectionParams::longterm_feerate`; `None` now means "use the target feerate", and the change policy is always waste-aware. The single value feeds both the change policy and the lowest-fee-bnb metric.
- `selection_algorithm_lowest_fee_bnb` no longer takes a `longterm_feerate` argument (it reads it from `SelectionContext`).
- Fix a false `CannotMeetTarget` under RBF: the reachability check now selects effective inputs at the resolved target feerate.

Removed:
- Remove the `Selector` struct (its lifecycle is folded into `InputCandidates::into_selection`).
- Remove `SelectorError` (replaced by `ChangePolicyError` and dedicated `IntoSelectionError` variants).
```

### Before submitting


src/input_candidates.rs
```diff
@@ -191,75 +191,176 @@ impl InputCandidates {
...
+
+        let mut cs = CoinSelector::new(self.coin_select_candidates());
+        if self.must_select().is_some() {
+            cs.select_next();
+        }
+
+        // Reachability pre-check.
+        {
+            let mut check = cs.clone();
+            check.select_all_effective(target.fee.rate);
```
noahjoeris, line 252: Is this pre-check right for RBF cases as well?

select_all_effective(target.fee.rate) only keeps EV > 0. under RBF, could we still need a rate-negative utxo to hit the absolute fee floor?

evanlinjin: Good point, there should be an absolute value check as well.

ValuedMammal:
> The public `Selector` struct is removed as nothing used it. Its methods were only ever called in sequence by `into_selection`, so it was public surface with no consumer.

I didn't understand the stated rationale. Public items are for consumers of the library, and not required to be used internally.

evanlinjin:
@ValuedMammal fair - that framing is incorrect and I'll fix it. Internal call sites aren't evidence about downstream consumers.

The actual rationale: the `Selector` was a wrapper that didn't provide much value. Algorithms now receive the downstream `CoinSelector` directly with a pure data `SelectionContext`. This is less code for us to maintain, and is actually more surface than what `CoinSelector` provided.

PR #85, opened 2026-07-14
Treat PSBT version as a minimum and auto-bump for CSV

### LINKED ISSUE #59: Remove `CreatePsbtError::UnsupportedVersion` variant
### Description

It is true that certain tx features require a minimum transaction version. Instead of erroring, we should just bump the transaction version.

This removes an error variant. It is the caller's responsibility to not pick inputs that require spending paths which require transaction versions that they do not wish to create.

### Proposed Changes

* Rename `PsbtParams::version` to `PsbtParams::min_version`.
* Tx version is bumped to `Version::TWO` if any input requires CSV. AFS also bumps the tx version if the sequence branch is picked.
* Update docs.
* Remove the `UnsupportedVersion` error in `CreatePsbtError`.

ANAMASGARD:
Closes #59
### Description

This PR fixes #59 by treating the transaction version in `PsbtParams` as a minimum version instead of an exact version.

Previously, transaction creation could fail when a selected input or anti-fee-sniping strategy required transaction version 2, but the caller requested version 1.

Now, the transaction version is automatically raised when required.

This PR:

* Renames `PsbtParams::version` to `PsbtParams::min_version`.
* Preserves the caller’s requested version when no feature requires a higher version.
* Raises the transaction version to `Version::TWO` when a selected input requires CSV.
* Raises the transaction version to `Version::TWO` when anti-fee-sniping selects the `nSequence` path.
* Preserves caller-provided versions higher than version 2.
* Removes `AntiFeeSnipingError::UnsupportedVersion`.
* Updates the documentation and changelog.
* Adds regression tests for the new behavior.

### Notes to the reviewers

`min_version` is treated only as a lower bound.

For example:

* `min_version: Version::ONE` with no CSV input remains version 1.
* `min_version: Version::ONE` with a CSV input becomes version 2.
* A version higher than 2 is not downgraded.
* The anti-fee-sniping `nLockTime` path does not unnecessarily raise the version.
* The anti-fee-sniping `nSequence` path raises the version to 2 when required.

The AFS sequence-path regression test uses a deterministic RNG so it does not depend on random branch selection.

### Changelog notice

* Rename `PsbtParams::version` to `PsbtParams::min_version`.
* Automatically raise the transaction version for CSV inputs and the AFS `nSequence` path.
* Remove the unsupported-version AFS error path.

### Verification

The following checks passed:

```text
cargo fmt --all -- --check
cargo test --all-features
cargo clippy --all-targets --all-features -- -D warnings
cargo check
cargo test --doc --all-features
RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps
git diff --check
```

`cargo check --no-default-features` currently fails inside `miniscript` because neither its `std` nor `no-std` feature is enabled. This appears to be an existing dependency-feature configuration issue and is unrelated to this change.

### Before submitting

* [x] I followed the contribution guidelines.
* [x] This PR breaks the existing API.

### DIFF
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -24,8 +24,11 @@ pub struct Selection {
 /// Parameters for creating a psbt.
 #[derive(Debug, Clone)]
 pub struct PsbtParams {
-    /// Use a specific [`transaction::Version`].
-    pub version: transaction::Version,
+    /// Minimum [`transaction::Version`] for the resulting transaction.
+    ///
+    /// The final version may be raised when selected inputs require CSV or when anti-fee-sniping
+    /// selects its `nSequence` strategy.
+    pub min_version: transaction::Version,
 
     /// Minimum tx locktime — a floor on the resulting `tx.lock_time`.
     ///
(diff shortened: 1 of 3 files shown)


src/selection.rs
```diff
@@ -766,28 +830,56 @@ mod tests {
...
-        let input = setup_test_input(confirmation_height).unwrap();
-        let inputs = vec![input];
-        let current_height = absolute::Height::from_consensus(confirmation_height + 50).unwrap();
+        let input = setup_test_input(confirmation_height)?;
+        let inputs = vec![input.clone()];
+        let current_height = absolute::Height::from_consensus(confirmation_height + 50)?;
 
         let mut tx = Transaction {
             version: Version::ONE,
             lock_time: LockTime::from_height(current_height.to_consensus_u32()).unwrap(),
```
aagbotemi: Test fails after rebasing on master.

AFS forces the nLockTime path whenever tx.lock_time != LockTime::ZERO. test_anti_fee_sniping_sequence_path_bumps_version builds its tx with a near-tip locktime, so after merging master it never reaches the nSequence branch:

```
assertion `left == right` failed
  left: Version(1)
  right: Version(2)
```
Building the tx with lock_time: LockTime::ZERO fixes it.

PR #87, opened 2026-08-20
fix(afs): require zero locktime for the nSequence path

evanlinjin:
### Description

Fixes https://github.com/bitcoindevkit/bdk-tx/pull/65#discussion_r3786557326 — point 2 of https://github.com/bitcoindevkit/bdk-tx/pull/73#issuecomment-5297469748.

The nSequence path exists so the tx resembles an off-chain settlement spending a timelock path, and those carry `nLockTime = 0` with a CSV-driven sequence.

#65 removed the `tx.lock_time = LockTime::ZERO` that path used to perform — correctly, since it regressed input CLTVs — but added no precondition in its place. So a tx whose locktime is already pinned, by an input's CLTV or by `min_locktime`, can come out with **both** a near-tip locktime and a confirmation-depth sequence. That matches neither an ordinary wallet spend nor a contract close: a third fingerprint, worse than either branch alone.

The path is only coherent at `lock_time == 0`, so make that a precondition:

```rust
let must_use_locktime = tx.lock_time != LockTime::ZERO
    || taproot_inputs.is_empty()
    || inputs.iter().any(|input| { /* ... */ });
```

### Changelog notice

```md
Fixed:
- Anti-fee-sniping only takes the `nSequence` path when `tx.lock_time` is zero. A locktime pinned by an input's CLTV or by `min_locktime` now always takes the `nLockTime` path.
```

### Before submitting

### DIFF
--- a/src/afs.rs
+++ b/src/afs.rs
@@ -117,7 +117,12 @@ pub(crate) fn apply_anti_fee_sniping(
         .collect();
 
     // Conditions that force nLockTime (vs nSequence).
-    let must_use_locktime = taproot_inputs.is_empty()
+    //
+    // The nSequence path exists so the tx resembles an off-chain settlement spending a timelock
+    // path, and those carry nLockTime = 0. So a locktime already pinned by an input's CLTV (or by
+    // `min_locktime`) rules the nSequence path out.
+    let must_use_locktime = tx.lock_time != LockTime::ZERO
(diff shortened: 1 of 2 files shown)


ValuedMammal (approved):
tACK e05146e87c4919145b42df82824fddf61b864e86

